PR #58, opened 2021-04-28
Ensure digest is same as chart (source) submitted

baijum:
When the has both chart and report. Ensure digest is the same as the
chart (source) submitted. The chart could be a tarball or in extracted
form.

Ref. https://issues.redhat.com/browse/HELM-173

### DIFF
--- a/scripts/src/chartprreview/chartprreview.py
+++ b/scripts/src/chartprreview/chartprreview.py
@@ -4,6 +4,8 @@ import sys
 import argparse
 import subprocess
 import json
+import hashlib
+import tempfile
 
 import requests
 import yaml
@@ -66,6 +68,19 @@ def verify_signature(category, organization, chart, version):
     out = subprocess.run(["gpg", "--verify", sign, report], capture_output=True)
     print(out.stdout.decode("utf-8"))
     print(out.stderr.decode("utf-8"))
+
+def match_checksum(category, organization, chart, version):
+    submitted_report_path = os.path.join("charts", category, organization, chart, version, "report.yaml")
+    submitted_report = yaml.load(open(submitted_report_path), Loader=Loader)
+    submitted_digest = submitted_report["metadata"]["tool"]["digest"]
+
+    generated_report_path = "report.yaml"
+    generated_report = yaml.load(open(generated_report_path), Loader=Loader)
+    generated_digest = generated_report["metadata"]["tool"]["digest"]
+
+    if  submitted_digest != generated_digest:
+        print("Digest is not matching:", submitted_digest, generated_digest)
+        sys.exit(1)
 
 def check_url(report_path):
     report = yaml.load(open(report_path), Loader=Loader)
@@ -164,7 +179,7 @@ def check_report_success(report_path, version):
         sys.exit(1)
 
 def generate_verify_report(category, organization, chart, version):
-    src = os.path.join("charts", category, organization, chart, version, "src")
+    src = os.path.join(os.getcwd(), "charts", category, organization, chart, version, "src")
     report_path = os.path.join("charts", category, organization, chart, version, "report.yaml")
     src_exists = False
     tar_exists = False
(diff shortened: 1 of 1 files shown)


mmulholla:
My suggestion for regenerating a new sha value to compare with the one from the submitted to report is to re-run the report against the chart and pull the sha from it, the run can specify ```-e has-readme``` so it only runs 1 check.
I have done this in a bash script.

pedjak (changes requested):
sharing the same view as @mmulholla

baijum:
Thank you @mmulholla and @pedjak for your suggestion. I have implemented the idea. If the user submits a source chart, a tarball will be generated using `helm package`, and then a report is generated from there. The digest is taken from there for comparison.

I have raised an issue with chart-verifier here: https://github.com/redhat-certification/chart-verifier/issues/92
Please take a look!

baijum:
Based on the comment here: https://github.com/redhat-certification/chart-verifier/issues/92#issuecomment-829198415
The logic has been changed to use the extracted source or tarball whichever is submitted by the user.

mmulholla:
In the new code I don't see where the new report is generated for getting the sha value.

scripts/src/chartprreview/chartprreview.py
```diff
@@ -181,13 +196,7 @@ def generate_verify_report(category, organization, chart, version):
...
             sys.exit(1)
     if src_exists:
-        out = subprocess.run(["helm", "package", src], capture_output=True)
-        stdout = out.stdout.decode("utf-8")
-        print(stdout)
-        print(out.stderr.decode("utf-8"))
-        dn = os.path.dirname(stdout.split(":")[1].strip())
-        fn = os.path.basename(stdout.split(":")[1].strip())
-        out = subprocess.run(["docker", "run", "-v", dn+":/charts:z", "--rm", "quay.io/redhat-certification/chart-verifier:latest", "verify", os.path.join("/charts/", fn)], capture_output=True)
+        out = subprocess.run(["docker", "run", "-v", src+":/charts:z", "--rm", "quay.io/redhat-certification/chart-verifier:latest", "verify", "/charts"], capture_output=True)
```
baijum, line 199: @mmulholla This is the line where the report is generated.

pedjak: but this line runs all checks - we thought only to run only digest computation.

mmulholla: If I understand the code correctly it generates a report whether not a report file was submitted, if a report was not submitted it serves as the invocation of the report, if not it is just used to check the sha.

The problem with this approach is we are going to install and test the chart even when we only want the report for the sha value.  

Also if we generate a report when one was not submitted it looks as though we don' save it anywhere, but should we not include it in the PR?

mmulholla: Just created a PR for my bash script. It includes an option  "checkdigest". More info available in the PR: https://github.com/openshift-helm-charts/repo/pull/63. Up to you whether you sue it or not.

baijum: @pedjak I have updated the command to only invoke "-e has-readme" check.

@mmulholla The "report.yaml" is saved in the current directory. And later it is used for matching name and the version is in the same as in the directory structure, submitted report and the chart (Chart.yaml) I would prefer not to invoke the docker command multiple times.  When the user submits a chart without a report, the same functions are reused.

mmulholla: Need to find some python lessons, sorry, now not sure how the second run does not overwrite the submitted one, just assure me that it will not.

baijum: That run is happening within an "if" condition like this: `if os.path.exists(report_path):` That condition becomes true when a user submits the report and the else is executed otherwise.

baijum: @mmulholla It will not overwrite the submitted one. The submitted report is in another directory.

mmulholla (approved):

Issue #66, opened 2021-05-01
Annotations are missing in the generated index.yaml


baijum:
PR #65 submitted a report without a chart. Here is the corresponding generated `index.yaml`: ddabaf7faef7d571f14a4249598ed30588150bc2
The `index.yaml` is missing the extra annotations from the script. Here is the "Release Charts" [job console output](https://github.com/openshift-helm-charts/repo/runs/2476784991#step:5:186):

```
2021-04-30T15:10:11.3876517Z [INFO] Creating Git worktree for index branch
2021-04-30T15:10:11.3877128Z Preparing worktree (detached HEAD 3821401)
2021-04-30T15:10:11.3877805Z HEAD is now at 3821401 Create CNAME
2021-04-30T15:10:11.3878120Z 
2021-04-30T15:10:11.3878527Z [INFO] Creating index from report
2021-04-30T15:10:11.3878997Z annotation {}
2021-04-30T15:10:11.3879387Z 
2021-04-30T15:10:11.3879738Z category: partners
2021-04-30T15:10:11.3880199Z Downloading index.yaml
2021-04-30T15:10:11.3880845Z [INFO] Updating the chart entry with new version
2021-04-30T15:10:11.3881381Z [INFO] Add and commit changes to git
2021-04-30T15:10:11.3881684Z 
2021-04-30T15:10:11.3882102Z [detached HEAD ddabaf7] Update index.html
2021-04-30T15:10:11.3882645Z  1 file changed, 60 insertions(+)
2021-04-30T15:10:11.3883109Z  create mode 100644 index.yaml
```

The annotation was empty (`{}`). The output is coming from the this command:
```
scripts/src/chartprreview/verify-report.sh annotations charts/partners/ibm/ibm-oms-ent-prod/6.0.0/report.yaml
```

baijum:
This has been addressed in #67

PR #67, opened 2021-05-01
Use sed to trim whitespace

baijum:
Fix issue #66

Based on the SO answer: https://stackoverflow.com/a/3232433/547840

### DIFF
--- a/scripts/src/chartprreview/verify-report.sh
+++ b/scripts/src/chartprreview/verify-report.sh
@@ -141,7 +141,7 @@ getAnnotations() {
 
 
   while IFS= read -r line; do
-    line=`echo $line | xargs`
+    line="$(echo -e "${line}" | sed -e 's/^[[:space:]]*//' -e 's/[[:space:]]*$//')"
     if [[ $line == "metadata:"* ]]; then
       metadata=true
     elif [[ $line == "results:"* ]]; then


scripts/src/chartprreview/verify-report.sh
```diff
@@ -141,7 +141,7 @@ getAnnotations() {
 
 
   while IFS= read -r line; do
-    line=`echo $line | xargs`
+    line="$(echo -e "${line}" | sed -e 's/^[[:space:]]*//' -e 's/[[:space:]]*$//')"
```
pedjak, line 144: we are reading here yaml structure, and reading it is line by line is very fragile and can catch things on e.g. wrong level. We should either use tools like [yq](https://github.com/mikefarah/yq) or perhaps this script should be written in Python or some other language with better yaml handling?

pedjak (changes requested):

scripts/src/chartprreview/verify-report.sh
baijum: I am fine with converting this script to Python. For now, let's merge this PR.

@mmulholla Any thoughts?

pedjak: ok, if this unblock you for a moment, but let's create an issue addressing this.

pedjak (approved):

scripts/src/chartprreview/verify-report.sh
baijum: I have created an issue here: #74

PR #104, opened 2021-05-07
Add note about making changes to a published chart

baijum:


docs/README.md
```diff
@@ -69,6 +69,10 @@ You can submit chart related changes in three methods:
 2. Submit a Chart Verification Report without the Chart
 3. Submit a Chart Verification Report with the Chart
 
+Making changes to a published chart version is not allowed.  If you notice any
+issue with the released version, you can create a new bugfix version and submit
+it.
+
```
pedjak, line 75: where does this requirement come from? is there any difficulties to allow this?

baijum: This requirement is not captured explicitly anywhere.  But I don't see a requirement for the update process defined anywhere. So, I thought of adding this to the document.

mmulholla: Maybe use: 
If you have an issue with a released chart please submit an updated bugfix version of the chart.

mmulholla: Agreed with Jasper to put a hold on this so closing the PR for now.

PR #112, opened 2021-05-07
Document missing OWNERS file error

baijum:
A follow-up PR for #110

PR #128, opened 2021-05-11
Consolidate all the jobs into a single pull_request_target based job

baijum:
Consolidate all the jobs into a single `pull_request_target` based job.
- Sanity Check
- Build and Verify
- Merge PR
- Release Charts (GitHub Release) and update `index.yaml`


.github/workflows/build.yml
```diff
@@ -1,15 +1,112 @@
...
+              repo: context.repo.repo,
+              issue_number: Number(issue_number),
+              name: 'authorized-request'
+            })
+
       - name: Checkout
         uses: actions/checkout@v2
+        with:
+          ref: ${{ github.event.pull_request.head.ref }}
+          repository: ${{ github.event.pull_request.head.repo.full_name }}
```
pedjak, line 109: we need to checkout this branch in a separate folder, to avoid overriding the target checkout and potential attacks. Perhaps we can do it manually, using `git`.

.github/workflows/build.yml
```diff
@@ -1,15 +1,112 @@
...
 
       - name: Save check run id
```
pedjak, line 111: this not needed anymore I think, because there is just single workflow now.

.github/workflows/build.yml
```diff
@@ -24,6 +121,7 @@ jobs:
           python-version: '3.9'
 
       - name: Verify PR
+        id: verify_pr
```
pedjak, line 124: this step can be moved under sanity check job, right?

.github/workflows/build.yml
```diff
@@ -37,8 +135,123 @@ jobs:
         if: always()
         run: |
           ve1/bin/pr-artifact --directory=./pr --pr-number=${{ github.event.number }} --api-url=${{ github.event.pull_request._links.self.href }}
+
       - uses: actions/upload-artifact@v2
         if: always()
         with:
           name: pr
           path: pr/
```
pedjak, line 143: IMHO, these two steps are not needed anymore.

.github/workflows/build.yml
```diff
@@ -37,8 +135,123 @@ jobs:
...
+        run: |
+          python3 scripts/prepare_pr_comment.py failure ${{ github.repository }}
+
+      - name: 'Prepare PR success comment'
+        if: ${{ steps.verify_pr.conclusion == 'success' }}
+        run: |
+          python3 scripts/prepare_pr_comment.py success ${{ github.repository }}
+
+      - name: 'Comment on PR'
+        if: always()
```
pedjak, line 156: is this line really needed?

.github/workflows/build.yml
```diff
@@ -37,8 +135,123 @@ jobs:
...
+            var comment = fs.readFileSync('./pr/comment', {encoding:'utf8', flag:'r'});
+            github.issues.createComment({
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              issue_number: Number(issue_number),
+              body: comment
+            });
+
+      - name: 'Add label to PR'
+        if: always()
```
pedjak, line 172: is this line really needed?

pedjak (changes requested):
let's collapse all jobs into single one, so that we avoid repeating steps (python install, for example) and make communication between steps easier.

.github/workflows/build.yml
```diff
@@ -1,15 +1,112 @@
...
+          python-version: '3.9'
+
+      - name: Sanity Check PR Content
+        id: sanity_check_pr_content
+        continue-on-error: true
+        run: |
+          python3 -m venv ve1
+          cd scripts && ../ve1/bin/pip3 install -r requirements.txt && cd ..
+          cd scripts && ../ve1/bin/python3 setup.py install && cd ..
+          ./ve1/bin/sanity-check-pr --api-url=${{ github.event.pull_request._links.self.href }}
```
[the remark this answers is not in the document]
baijum: I think we can keep sanity very minimal. There are more checks about `OWNERS` file and its content later in the `Build and Verify` section.
1. https://github.com/openshift-helm-charts/charts/tree/main/docs#pull-request-author-is-not-part-of-owners-file
2. https://github.com/openshift-helm-charts/charts/pull/112 (waiting for approval)

.github/workflows/build.yml
baijum: Removed. 76f652c5dcc1eb7d3443167b1c25c64c3fade50c

.github/workflows/build.yml
baijum: Removed. 76f652c5dcc1eb7d3443167b1c25c64c3fade50c

.github/workflows/build.yml
```diff
@@ -37,8 +128,121 @@ jobs:
...
+          GITHUB_REF: ${{ github.ref }}
+          GITHUB_TOKEN: ${{secrets.GITHUB_TOKEN}}
+        id: release-charts
+        run: |
+          curl -L -o cr.tar.gz https://github.com/helm/chart-releaser/releases/download/v1.2.0/chart-releaser_1.2.0_linux_amd64.tar.gz
+          tar zxvf cr.tar.gz
+          sudo cp -f cr /usr/local/bin/cr
+          python3 -m venv ve1
+          cd scripts && ../ve1/bin/pip3 install -r requirements.txt && cd ..
+          docker pull quay.io/redhat-certification/chart-verifier:latest
```
baijum, line 234: Probably this docker pull is not required here.

.github/workflows/build.yml
baijum: Now everything under a common job.

.github/workflows/build.yml
```diff
@@ -37,8 +117,115 @@ jobs:
...
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Dump GitHub context
+        env:
+          GITHUB_CONTEXT: ${{ toJson(github) }}
+        run: |
+          echo "$GITHUB_CONTEXT"
+
+      - name: Configure Git
```
mmulholla, line 201: Do we still need this step?

mmulholla, line 201: Do we need this step?

baijum: Git commit will not work without this. Also, we need a way to identify who made the commit when looking at the git history.

mmulholla:
@baijum @pedjak 
I am good to merge this and continue to make adjustements as needed.

baijum:
> @baijum @pedjak
> I am good to merge this and continue to make adjustments as needed. It makes significant progress and auto-merge works.

I am fine with that approach. We can continue refactoring after this PR is merged. This will be helpful in my chart-testing PR.

mmulholla:
Fixed a couple of insert errors in messages and collapsed to one commit. Looking good.

PR #145, opened 2021-05-13
Remove unused workflows and scripts

baijum:
Now all the jobs are consolidated into a single
pull_request_target based job.

Ref. PR #128


baijum:
Since `pull_request_target` is used there is no need to manually approve any PRs from new submitters.

PR #146, opened 2021-05-13
Fix parsing error and make chart-testing mndatory

mmulholla:
A few changes:
- Fixed problem if reason contains a ":" which was causing missing quotes
- made chart-testing mandatory
- changed the annotation  namespace to  charts.openshift.io
- removed the now redundant code allowing for minkubeverion and lastCertifiedTime 
- included certifiedOpenShiftVersions in annotations


dperaza4dustbit (approved):
Looks good

baijum (approved):
LGTM

PR #152, opened 2021-05-14
Fix location of index.yaml to publish

baijum:
import missing sys

fix shell syntax


.github/workflows/build.yml
```diff
@@ -245,7 +245,7 @@ jobs:
         continue-on-error: true
         with:
           tag_name: ${{ steps.release-charts.outputs.tag }}
-          files: report.yaml
+          files: pr-branch/report.yaml
```
baijum, line 248: This is required to publish `report.yaml` as a GitHub release along with the chart tarball.

dperaza4dustbit: I see

dperaza4dustbit (approved):
+1

PR #165, opened 2021-05-14
Gh pages commit msg

mmulholla:
Changed commit message to: 
```<partner>-<chart>-<version> Index.yaml```

also fixed an error in verify_pr where the varibale was wrong resulting in oc-Login happening even when report was included in submission.


baijum:
These commits were part of PR #166

PR #167, opened 2021-05-15
Validate OpenShift version and trim to MAJOR.MINOR format

baijum:
Ensure the version has atleast MAJOR.MINOR versions as part of the version.
Save MAJOR.MINOR version to to be part of `index.yaml`
Ref. https://github.com/redhat-certification/chart-verifier/issues/111

### DIFF
--- a/scripts/src/chartprreview/chartprreview.py
+++ b/scripts/src/chartprreview/chartprreview.py
@@ -7,6 +7,7 @@ import json
 import hashlib
 import tempfile
 
+import semver
 import requests
 import yaml
 try:
@@ -220,6 +221,13 @@ def check_report_success(directory, report_path, version):
         msg = f"[ERROR] Missing annotation in chart/report: {annotation}"
         write_error_log(directory, msg)
         sys.exit(1)
+
+    if "charts.openshift.io/certifiedOpenShiftVersions" in annotations:
+        full_version = annotations["charts.openshift.io/certifiedOpenShiftVersions"]
+        if not semver.VersionInfo.isvalid(full_version):
+            msg = f"[ERROR] OpenShift version not conforming to SemVer spec: {full_version}"
+            write_error_log(directory, msg)
+            sys.exit(1)
 
     out = subprocess.run(["scripts/src/chartprreview/verify-report.sh", "results", report_path], capture_output=True)
     r = out.stdout.decode("utf-8")
--- a/scripts/src/chartrepomanager/chartrepomanager.py
+++ b/scripts/src/chartrepomanager/chartrepomanager.py
@@ -4,12 +4,12 @@ import os
 import sys
 import re
 import subprocess
-import datetime
 import tempfile
 from datetime import datetime, timezone
 import json
 import urllib.parse
 
+import semver
 import requests
 import yaml
 try:
@@ -245,6 +245,11 @@ def update_chart_annotation(category, organization, chart_file_name, chart, repo
         out = yaml.load(data, Loader=Loader)
         vendor_name = out["vendor"]["name"]
         annotations["charts.openshift.io/provider"] = vendor_name
+
+    if "charts.openshift.io/certifiedOpenShiftVersions" in annotations:
+        full_version = annotations["charts.openshift.io/certifiedOpenShiftVersions"]
+        ver = semver.VersionInfo.parse(full_version)
+        annotations["charts.openshift.io/certifiedOpenShiftVersions"] = f"{ver.major}.{ver.minor}"
 
     out = subprocess.run(["tar", "zxvf", os.path.join(".cr-release-packages", f"{organization}-{chart_file_name}"), "-C", dr], capture_output=True)
     print(out.stdout.decode("utf-8"))
--- a/scripts/requirements.txt
+++ b/scripts/requirements.txt
@@ -3,4 +3,5 @@ chardet==4.0.0
 idna==2.10
 PyYAML==5.4.1
 requests==2.25.1
+semver==2.13.0
 urllib3==1.26.4
--- a/scripts/setup.cfg
+++ b/scripts/setup.cfg
@@ -26,6 +26,7 @@ python_requires = >=3.6
 install_requires =
     PyYAML
     requests
+    semver
 
 [options.packages.find]
 where = src


baijum:
This can be merged after https://github.com/redhat-certification/chart-verifier/pull/115

baijum:
> This can be merged after [redhat-certification/chart-verifier#115](https://github.com/redhat-certification/chart-verifier/pull/115)

The above PR is merged.

mmulholla:
Do we need the check in the release process as we would not have got this far had the OC Version no been valid?

baijum:
> Do we need the check in the release process as we would not have got this far had the OC Version no been valid?

Removed that extra error case during release.

baijum:
Let's hold this PR until #170 is merged.

scripts/src/chartprreview/chartprreview.py
```diff
@@ -221,6 +221,15 @@ def check_report_success(directory, report_path, version):
         write_error_log(directory, msg)
         sys.exit(1)
 
+    if "charts.openshift.io/certifiedOpenShiftVersions" in annotations:
+        pattern = re.compile("(\d+).(\d+).*")
```
pedjak, line 225: not sure if the reg is correct - `.` matches anything not only dots.

there are plenty of libraries that can be used for parsing versions - for example: https://pypi.org/project/semver/ -

baijum: The input is coming from the chart-verifier tool.

1. It could be either an output of [oc version command](https://github.com/redhat-certification/chart-verifier/blob/511119c1af7081392bf40f440debc20948268983/pkg/tool/oc.go#L23)
2. Alternative could be a direct [input from the user which is validated for SemVer spec](https://github.com/redhat-certification/chart-verifier/blob/511119c1af7081392bf40f440debc20948268983/pkg/chartverifier/certifier.go#L88).

In this case, we are interested in the `MAJOR.MINOR` part of the version to set as an annotation.  So, I don't see a need to verify the version for SemVer spec again.

pedjak: We want to check if the value is according to semver format, and the current code is using regex pattern that matches strings that are not semver versions, take a look at this:

```python
mport re

v = "1a2"

pattern = re.compile("(\d+).(\d+).*")
match = pattern.match(v)
if match:
    print("matched")
```

Let's not reinvent the wheel here - instead we can use some library for parsing the version and implicitly validating it. https://pypi.org/project/semver/ is just one of them

scripts/src/chartrepomanager/chartrepomanager.py
```diff
@@ -246,6 +246,13 @@ def update_chart_annotation(category, organization, chart_file_name, chart, repo
         vendor_name = out["vendor"]["name"]
         annotations["charts.openshift.io/provider"] = vendor_name
 
+    if "charts.openshift.io/certifiedOpenShiftVersions" in annotations:
+        pattern = re.compile("(\d+).(\d+).*")
+        full_version = annotations["charts.openshift.io/certifiedOpenShiftVersions"]
+        match = pattern.match(full_version)
+        if match:
+            annotations["charts.openshift.io/certifiedOpenShiftVersions"] = ".".join(match.groups())
+
```
pedjak, line 255: see above comment.

scripts/src/chartprreview/chartprreview.py
baijum: Done. 89b9639d6d5a12a7c6985ee59479594fabf69a31

scripts/src/chartrepomanager/chartrepomanager.py
baijum: Done. 89b9639d6d5a12a7c6985ee59479594fabf69a31

baijum:
I have tested this PR here: https://github.com/miles2go/repo/pull/73

To test this, I cherry-picked changes from #170 and it passed all tests and PR got merged.
See the index.yaml `charts.openshift.io/certifiedOpenShiftVersions` annotation:
https://github.com/miles2go/repo/blob/32bcfa5f3205a569105cedc6fbe3476ce055f5a0/index.yaml#L121

scripts/src/chartrepomanager/chartrepomanager.py
```diff
@@ -246,6 +246,11 @@ def update_chart_annotation(category, organization, chart_file_name, chart, repo
         vendor_name = out["vendor"]["name"]
         annotations["charts.openshift.io/provider"] = vendor_name
 
+    if "charts.openshift.io/certifiedOpenShiftVersions" in annotations:
+        full_version = annotations["charts.openshift.io/certifiedOpenShiftVersions"]
+        ver = semver.VersionInfo.parse(full_version)
+        annotations["charts.openshift.io/certifiedOpenShiftVersions"] = ".".join((str(ver.major), str(ver.minor)))
```
pedjak, line 252: IMHO, `annotations["charts.openshift.io/certifiedOpenShiftVersions"] = f"{ver.major}.{ver.minor}"` is more readable, i.e. the intention is clearer.

scripts/src/chartprreview/chartprreview.py
```diff
@@ -221,6 +222,13 @@ def check_report_success(directory, report_path, version):
         write_error_log(directory, msg)
         sys.exit(1)
 
+    if "charts.openshift.io/certifiedOpenShiftVersions" in annotations:
+        full_version = annotations["charts.openshift.io/certifiedOpenShiftVersions"]
+        if not semver.VersionInfo.isvalid(full_version):
+            msg = f"[ERROR] No matching OpenShift version found: {full_version}"
```
pedjak, line 228: IMHO, the message could be improved, because we land here if the `full_version` is not a valid sem version.

baijum: Done. 908e8605354d7c73c1171c77f37b9ffeee336197

scripts/src/chartrepomanager/chartrepomanager.py
baijum: Done. 908e8605354d7c73c1171c77f37b9ffeee336197

pedjak (approved):

PR #168, opened 2021-05-15
Ensure namespace and SA are created

baijum:
The newly created namespace and service account are not available for login and switching the context immediatly. There was an explicit sleep introduced in PR #168 as a workaround to wait for resource availability. This PR introduces an incremental check for resource availability. Once all the resources are realized, the program exit normally.

### DIFF
--- a/scripts/src/saforcharttesting/saforcharttesting.py
+++ b/scripts/src/saforcharttesting/saforcharttesting.py
@@ -214,6 +214,27 @@ def write_sa_token(namespace, token):
                 with open(token, "w") as fd:
                     fd.write(base64.b64decode(content).decode("utf-8"))
 
+def switch_project_context(namespace, token):
+    tkn = open(token).read()
+    for i in range(7):
+        out = subprocess.run(["./oc", "login", "--token", tkn, "--server", "https://api.ocpappsvc-osd.zn6c.p1.openshiftapps.com:6443"], capture_output=True)
+        stdout = out.stdout.decode("utf-8")
+        print(stdout)
+        out = subprocess.run(["./oc", "project", namespace], capture_output=True)
+        stdout = out.stdout.decode("utf-8")
+        print(stdout)
+        out = subprocess.run(["./oc", "config", "current-context"], capture_output=True)
+        stdout = out.stdout.decode("utf-8").strip()
+        print(stdout)
+        if stdout.endswith(":".join((namespace, namespace))):
+            print("current-context:", stdout)
+            return
+        time.sleep(10)
+
+    # This exit will happen if there is an infra failure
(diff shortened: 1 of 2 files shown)


mmulholla:
We now seem to co oc_login twice? Is that needed?

baijum:
> We now seem to do oc_login twice? Is that needed?

Removed extra login.

> Also please provide a description of what the PR is doing.

Done.

baijum:
I have tested this PR here: https://github.com/miles2go/repo/pull/70

If no other concern, let's merge this PR.

baijum:
This change will make PR #170 in conflict. So, let's not merge this PR.

PR #170 has more priority.

scripts/src/saforcharttesting/saforcharttesting.py
```diff
@@ -156,6 +156,23 @@ def write_sa_token(namespace, token):
                 with open(token, "w") as fd:
                     fd.write(base64.b64decode(content).decode("utf-8"))
 
+def switch_project_context(namespace, token):
+    tkn = open(token).read()
+    for i in range(7):
```
pedjak, line 161: why do we set limit to 7 iterations? How do we know this is enough and what happens if it is not?

baijum: I noticed the resource becomes available within 1 minute. Currently, the maximum waiting time is 7x10 = 70 seconds.
I think we can exit the program with error code 1. so that the job is going to fail.  This should be considered an infra failure. I will write a detailed error message to help the person who reads it.

baijum: Added program exit with error message: 89ef26372c26613c564f3613ba19fa78eb85fa49

baijum:
This PR is not mandatory for GA. The workaround in PR #168 seems to be effective so far.

scripts/src/saforcharttesting/saforcharttesting.py
```diff
@@ -214,6 +214,28 @@ def write_sa_token(namespace, token):
...
+        stdout = out.stdout.decode("utf-8").strip()
+        print(stdout)
+        if stdout.endswith(":".join((namespace, namespace))):
+            print("current-context:", stdout)
+            return
+        time.sleep(10)
+
+    # This exit will happen if there is an infra failure
+    print("""[ERROR] There is an error creating namespace and service account.
+This is an infrastructure failure.  It is not directly related to the changes in the pull request.""")
```
pedjak, line 236: Would it help to use to tell him to retry the job?

baijum: Yes. See this change. db4d14e505780f59c6acb74310b542caff31dfd0

pedjak (approved):

PR #175, opened 2021-05-18
Set the right expection for PGP key and signature

baijum:
Both public key and signature must be conforming to ASCII armor format.


docs/README.md
```diff
@@ -161,11 +161,13 @@ file name as `report.yaml`.
 
 When you follow the [partner connect documentation][partners], you could see
 details about adding a PGP public key.  Adding PGP public key is optional.  But
-if you have added that, you should see your public key in the _OWNERS_ file under
-your chart directory within your organization directory.
+if you have added that, you should see your public key in the _OWNERS_ file
+under your chart directory within your organization directory.  The PGP public
+key is available in the `publicPgpKey` attribute.  The value of that attribute
+must be conforming to [ASCII armor format][ascii-armor].
```
pedjak, line 167: can we provide some link here showing how to generate key in such format? The current link is very academic.

baijum: Replaced the link with a Red Hat article: https://www.redhat.com/sysadmin/creating-gpg-keypairs

pedjak (approved):

PR #177, opened 2021-05-19
Ensure the job is failed for PR merge failure

### LINKED ISSUE #176: Update chart issue
I created a chart using src and a report.

then used a new branch to merge a one line change to the Chart.yaml and new report, I had merge issues so not automerge but workflow worked and released the chart before resolving merge issues.

did another update without merge problems and everything worked and chart was released.  

In both cases Index.yaml was updated with new values.

Updating a chart like t his should be a worklflow error

baijum:
* Release chart if merge is a success

Fixes #176


mmulholla:
#176 is not fully addressed by this PR, the other problem is that we update an existing version in the Index.yaml.

baijum:
> #176 is not fully addressed by this PR, the other problem is that we update an existing version in the Index.yaml.

So, if the version already exists in `index.yaml` we should not fail with an explanation, right?

baijum:
> > #176 is not fully addressed by this PR, the other problem is that we update an existing version in the Index.yaml.
> 
> So, if the version already exists in `index.yaml` we should fail with an explanation, right?
> 
> EDIT: s/should not fail/should fail

Based on the engineering call discussion today we need to allow updating the `index.yaml`.

There is another requirement to allow replacing the released chart itself. I will create a separate story for the same.

baijum:
I have created a PR in sandbox repo to demonstrate this PR: https://github.com/openshift-helm-charts/sandbox/pull/6
The `dev` branch in that repository has the changes in this PR.

.github/workflows/build.yml
```diff
@@ -173,7 +174,7 @@ jobs:
                 issue_number: Number(issue_number),
                 owner: context.repo.owner,
                 repo: context.repo.repo,
-                labels: ['authorized-request']
+                labels: ['authorized-request', 'automerge']
```
pedjak, line 177: what is the semantic of new `automerge` label? and how is different from `sanity-ok`? this steps will be executed  under the same condition as the step in line 31: https://github.com/openshift-helm-charts/charts/pull/177/files#diff-5c3fa597431eda03ac3339ae6bf7f05e1a50d6fc7333679ec38e21b337cb6721R31 IMHO, both labels have the same semantic, Hence, I would not introduce `automerge`.  Furthermore I would even argue that `sanity-ok` label is as well not needed - `authorized-request` label is added when sanity check is successful - so they have the same semantic.

What I miss however is a label (e.g. `valid-chart-pr`) stating that chart checks are passing - and that should come as the outcome of `verify-pr` step. With that, PR should merged only if both `authorized-request` and `valid-chart-pr` labels are present.

cc @mmulholla - would like to hear your opinion.

.github/workflows/build.yml
```diff
@@ -190,7 +191,8 @@ jobs:
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
           MERGE_METHOD: squash
-          MERGE_LABELS: ""
+          MERGE_LABELS: "automerge,!hold"
+          MERGE_REMOVE_LABELS: "automerge"
```
pedjak, line 195: IMHO it would be simple not to reach this step at all, i.e. not approve PR and fail the job. How about to have a step before "Approve PR" like the following:

```
- if: ${{ steps.verify_pr.conclusion == 'failure }}
  run: exit 1
```

With that we do no need "Check for automerge" step,  nor any condition on "Merge PR" step.

pedjak (changes requested):

.github/workflows/build.yml
baijum: I did this change but seems to not working in the case of conflict. So, I am going to revert this change unless I missed something. 
Here is the test PR: https://github.com/openshift-helm-charts/sandbox/pull/7

pedjak: It is because `verify_pr` step never fails - the last command executed there is `cd ..` and it is always successful.

baijum: There are scenarios where `verify_pr` fails. The Python script calls `sys.exit(1)` whenever there is an issue with checks.
In the above [test PR](https://github.com/openshift-helm-charts/sandbox/pull/7) all checks were passing.  But the merge didn't happen because there is a conflict.  So, we cannot depend on the failure of `verify_pr` step to decide to make the chart release.

pedjak: > There are scenarios where verify_pr fails. The Python script calls sys.exit(1) whenever there is an issue with checks.

My bad, the script is really configured to fail fast, I missed that.

However,  the automerge action does not fail if merge cannot be done due to conflicts. A way to check this would be to check the PR status via something like this:

```
var pr = github.pulls.get({
      owner: context.repo.owner,
      repo: context.repo.repo,
      pull_number: ${{ github.event.pull_request.number }}
     });
if (!pr.merged) {
       core.setFailed('not merged')
}
```

And this should happen probably in some sort of loop. Checking for removed label is just not enough - there could be still a room for failing the merge.

baijum:
I have moved draft PR related part to a separate PR #186

.github/workflows/build.yml
baijum: Done. df8a9de9cae73b7d95e2477398200d5e20e23278

.github/workflows/build.yml
```diff
@@ -202,6 +202,10 @@ jobs:
           git config --global user.name "github-actions[bot]"
           git config --global user.email "41898282+github-actions[bot]@users.noreply.github.com"
 
+      - name: Check for PR merge
+        run: |
+          ./ve1/bin/check-auto-merge --api-url=${{ github.event.pull_request._links.self.href }}
+
```
pedjak, line 208: Should not be better if check this right after `merge_pr` step so that we can fail fast? 

Also, any advantages of using pythin here vs. github-script action?

pedjak (approved):

Issue #182, opened 2021-05-20
Impacts of updating the existing released version of a Helm chart with the same version number


baijum:
A partner can not update the released version of a Helm chart with the same version number. This issue explains the impacts of updating the existing released version of a Helm chart with the same version number.

1.  When the Helm chart is getting published through the repository, it will reach the end-users immediately. And if a modified version publishes with the same version number, the end-user will not receive the changes. The version deployed on different environments (Dev, QA, Stage, Prod, etc.) will not be the same. It is going to make identifying issues very difficult.

2. Helm charts best practices document recommend using SemVer spec for versioning. Ref. https://helm.sh/docs/chart_best_practices/conventions/#version-numbers Version number increment for changes is a fundamental principle embraced by the SemVer spec.

3. It is nearly impossible to give security advisories without proper versioning of Helm chart releases.  If there is a Helm chart with different changesets, assigning a CVE entry would be difficult.

baijum:
A related PR: #104 (Based on the discussion here, we can reopen it)

dperaza4dustbit:
If no objections are raised I would fail on an attempt to change existing charts with a very clear message asking to re-submit

mmulholla:
Would I workflow also fail as the chart tag and release would already exist?

PR #183, opened 2021-05-20
Prevent updating existing releases

baijum:
Prevent a partner from updating the released version of a Helm chart with the
same version number. Ref. #182

Jira: https://issues.redhat.com/browse/HELM-186

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -21,11 +21,14 @@ jobs:
       - name: Sanity Check PR Content
         id: sanity_check_pr_content
         continue-on-error: true
+        env:
+          GITHUB_REF: ${{ github.ref }}
         run: |
           python3 -m venv ve1
           cd scripts && ../ve1/bin/pip3 install -r requirements.txt && cd ..
           cd scripts && ../ve1/bin/python3 setup.py install && cd ..
-          ./ve1/bin/sanity-check-pr --api-url=${{ github.event.pull_request._links.self.href }}
+          INDEX_BRANCH=$(if [ "${GITHUB_REF}" = "refs/heads/main" ]; then echo "refs/heads/gh-pages"; else echo "${GITHUB_REF}-gh-pages"; fi)
+          ./ve1/bin/sanity-check-pr --index-branch=${INDEX_BRANCH} --repository=${{ github.repository }} --api-url=${{ github.event.pull_request._links.self.href }}
 
       - uses: actions/github-script@v3
         if: ${{ steps.sanity_check_pr_content.outcome == 'success'}}
@@ -140,6 +143,8 @@ jobs:
 
       - name: 'Prepare PR comment'
         if: always()
+        env:
+          SANITY_ERROR_MESSAGE: ${{ steps.sanity_check_pr_content.outputs.sanity-error-message }}
         run: |
           python3 scripts/prepare_pr_comment.py ${{ steps.sanity_check_pr_content.outcome }} ${{ steps.verify_pr.conclusion }} ${{ github.repository }}
 
@@ -212,14 +217,18 @@ jobs:
           GITHUB_REF: ${{ github.ref }}
           GITHUB_TOKEN: ${{secrets.GITHUB_TOKEN}}
           REPORT_CONTENT: ${{steps.verify_pr.outputs.report_content}}
+          CHART_ENTRY_NAME: ${{ steps.sanity_check_pr_content.outputs.chart-entry-name }}
+          CHART_NAME_WITH_VERSION: ${{ steps.sanity_check_pr_content.outputs.chart-name-with-version }}
         id: release-charts
         run: |
           curl -L -o cr.tar.gz https://github.com/helm/chart-releaser/releases/download/v1.2.0/chart-releaser_1.2.0_linux_amd64.tar.gz
           tar zxvf cr.tar.gz
           sudo cp -f cr /usr/local/bin/cr
           INDEX_BRANCH=$(if [ "${GITHUB_REF}" = "refs/heads/main" ]; then echo "refs/heads/gh-pages"; else echo "${GITHUB_REF}-gh-pages"; fi)
+          CWD=`pwd`
           cd pr-branch
           ../ve1/bin/chart-repo-manager --repository=${{ github.repository }} --index-branch=${INDEX_BRANCH} --api-url=${{ github.event.pull_request._links.self.href }} --pr-number=${{ github.event.number }}
+          cd ${CWD}
 
       - name: Release
         uses: softprops/action-gh-release@master
--- a/docs/README.md
+++ b/docs/README.md
@@ -225,23 +225,24 @@ be merged automatically.
 
 If there is an issue, a comment with details will appear in the pull request.
 At the bottom of the page, you will also see a box with a heading like this:
-"Some checks were not successful".  And below that, two checks are marked with
-the "required" label to merge the PR automatically:
+"Some checks were not successful" or "All checks have failed".  And below that,
+checks are marked with the "required" label to merge the PR automatically:
 
-1. Sanity Check / sanity-check (pull_request_target)
-2. Build and Verify / build-and-verify (pull_request)
+> CI / Chart Certification (pull_request_target)            [Details](#)
 
-The first one, Sanity Check, fails if the pull request contains changes not
-related to any chart.  This check performs using the script already part of the
-upstream repository.  That way, it will ensure the pull request author is not
-modifying the verification scripts themselves.
+The first part of Chart Certification (Sanity Check) fails if the pull request
+contains changes not related to any chart.  This check performs using the script
+already part of the upstream repository.  That way, it will ensure the pull
+request author is not modifying the verification scripts themselves.  Sanity
+check also ensures the chart version submitted is not already released
+earlier. The job will fail if any sanity checks fail.
 
-The second job, Build and Verify, runs within an unprivileged environment (for
-[security reasons][unprivileged-environment]) and performs all the necessary
+The second part of Chart Certification (Verify PR) performs all the necessary
 checks to merge the pull request.  The following sections explain various errors
-produced by the Build and Verify job.
+produced by the Verify PR.
 
-*Note*: Please ignore other infra jobs like Enable Auto Merge, Check labels, etc.
+The third part of Chart Certification makes the chart release and update
+`index.yaml` in the repository.
 
 ### Error when submitting files not part of any chart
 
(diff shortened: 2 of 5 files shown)


scripts/src/sanitycheckpr/sanitycheckpr.py
```diff
@@ -17,23 +23,69 @@ def ensure_only_chart_is_modified(api_url):
...
+        if not match:
+            msg = f"[ERROR] PR should only modify chart related files: {filename}"
+            print(msg)
+            print(f"::set-output name=sanity-error-message::{msg}")
             sys.exit(1)
-        elif reportpattern.match(f["filename"]):
+        elif match:
+            if not match_found:
+                pattern_match = match
+                match_found = True
```
mmulholla, line 41: "elif match" should be "else" 
given match is True "pattern_match=True" is easier to follow.

scripts/src/sanitycheckpr/sanitycheckpr.py
```diff
@@ -17,23 +23,69 @@ def ensure_only_chart_is_modified(api_url):
...
+        entry_name = f"{organization}-{chart}"
+        d = data["entries"].get(entry_name, [])
+        for v in d:
+            if v["version"] == version:
+                msg = f"[ERROR] Helm chart release already exists in the index.yaml: {version}"
+                print(msg)
+                print(f"::set-output name=sanity-error-message::{msg}")
+                sys.exit(1)
+
+        tag_name = f"{organization}-{chart}-{version}"
```
mmulholla, line 67: tag_name is also determined in chartrepomanger.py, as a result it would be better to set tag_name as an output which chartrepomanger.py can use. This way we ensure both are using the same thing and we don't need to update two places if it ever changed for any reason. The same goes for entry_name.

scripts/src/sanitycheckpr/sanitycheckpr.py
```diff
@@ -17,23 +23,69 @@ def ensure_only_chart_is_modified(api_url):
...
+        tag_api = f"https://api.github.com/repos/{repository}/git/ref/tags/{tag_name}"
+        headers = {'Accept': 'application/vnd.github.v3+json'}
+        print(f"[INFO] checking tag: {tag_api}")
+        r = requests.head(tag_api, headers=headers)
+        if r.status_code == 200:
```
mmulholla, line 72: To get here there is no entry for the chart in Index.yaml, so could we ever get a chart which has not entry but does have a release? If not I am not sure we need this check.

baijum: A tag cannot exist without a corresponding entry in the `index.yaml` in an ideal condition. Since tags can be created by other means, I thought it is better to safeguard against tag existence.

scripts/src/sanitycheckpr/sanitycheckpr.py
baijum: Done. bf4863c47c0f01ad2f4bfe80a388d432e6d621b9

scripts/src/sanitycheckpr/sanitycheckpr.py
baijum: Done. bf4863c47c0f01ad2f4bfe80a388d432e6d621b9

scripts/src/sanitycheckpr/sanitycheckpr.py
```diff
@@ -17,23 +23,71 @@ def ensure_only_chart_is_modified(api_url):
     files_api_url = f'{api_url}/files'
     headers = {'Accept': 'application/vnd.github.v3+json'}
     r = requests.get(files_api_url, headers=headers)
-    pattern = re.compile(r"charts/(\w+)/([\w-]+)/([\w-]+)/([\w\.]+)/.*")
+    pattern = re.compile(r"charts/(\w+)/([\w-]+)/([\w-]+)/(.+?(?=\/))/.*")
```
pedjak, line 26: the current and proposed code changes still make possible that a PR contains changes to multiple charts. Namely, the regex only looks if filepaths are starting with `charts/` prefix and have the defined depth.

IMHO, forbidding updates can be done a way simpler, even without checking index.yaml:

* Get list of files submitted in PR (that is already done in line 25)
* All file statuses should be set to `added` - otherwise, we can exit with error - if PR contains some files that are modified, it means that PR modifies some existing chat, and that chart is then already published, because it is a part of code repo.
* All files in the PR should have share the same common prefix

There is no need to download any `index.yaml` and the whole workflow logic becomes a way simpler.

pedjak: cc @mmulholla

baijum: Done. 5d5ac7e4ef53852d1935ce45c92b993980b7884d

mmulholla: It is possible to update a chart by adding a new file,  which this would now miss. Not sure if you can add a file which could affect the contents of the Index.yaml file but we would continue to create the release. The previous approach required more code to cover this issue.

mmulholla: maybe we just add a check that Chart.yaml is included in the PR?

baijum: Added checking the existence of chart and/or report as part of the PR:
0f8447bdd008cb36e49214fbf310db4e69feea04

docs/README.md
```diff
@@ -196,23 +196,24 @@ be merged automatically.
 
 If there is an issue, a comment with details will appear in the pull request.
 At the bottom of the page, you will also see a box with a heading like this:
-"Some checks were not successful".  And below that, two checks are marked with
-the "required" label to merge the PR automatically:
+"Some checks were not successful" or "All checks have failed".  And below that,
```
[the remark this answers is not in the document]
baijum: See the status part of this PR. Currently, it's written, "All checks have failed". But as soon as we introduce one more check, the message could be either "Some checks were not successful" or "All checks have failed".
This is a GitHub UX.

scripts/src/sanitycheckpr/sanitycheckpr.py
mmulholla: I don't think  the logic is quite right yet.

report found, need to ensure no other files are added unless chart.yaml is present.  
report found, tarball found, need to ensure no other files are added.
no report, must ensure either chart.yaml is included, or ,tgz is the only file.

baijum: These combination checks are present as part of `Verify PR - generate report`  step.
In the sanity check, we need to ensure only addition is happening. And you pointed what if the user adds an extra file to the existing chart. To last commit ensure "Chart.yaml" is in the add "added" list (Same with report and chart).  This was we can ensure only addition is part of the PR. And if it's all addition there is an addition to one of the above three files.

mmulholla: I thought the idea here was to prevent updating an existing chart by detection in sanity check.
1. I can publish a chart without a report and then add a report later, this should be caught but is not in the current logic. 
2. I can publish a chart using just a report, then add chart source or a tarball, this should be caught but is not in the current logic.   
3. Not sure what it means if I publish a chart using source and then try adding a tarball, but should probably also be caught in this logic.

Beginning to think this concept of making sure all contents of a PR are additions is flawed, ad the old logic was better.

baijum: @mmulholla I agree that this logic is flawed! I am with going back to the old approach in PR.
@pedjak Any comments?

pedjak: good catch @mmulholla 

In order to fix the mentioned issues, we should add one more check:

* the chart directory must not exist in the target branch

baijum: @mmulholla @pedjak @dperaza4dustbit This discussion is going back and forth multiple times. I need a confirmation whether I should revert my changes as suggested by @mmulholla or try the last suggestion given by @pedjak .

mmulholla: Personally I prefer checking Index.yaml content because we are checking exactly what we are trying to avoid, rather than checking some upstream status only to find some corner case that it does not cover.  Checking the chart version directory does not already exist certainly makes for simpler code but I would worry that we could merge a chart and then fail in the release process for some reason. In this case one solution we would now prevent is  to trigger the release process again using a new PR with a irrelevant change.  I am assuming that chart content which passes build and verify could never cause a release proces failure.

mmulholla: @predrag please respond to my comment. Will leave the final decision to you.

pedjak: > please respond to my comment. Will leave the final decision to you.

I am fine if you wanna take this route.

dperaza4dustbit: Why don't we define a set of test cases we want to pass to call this PR done. Then we check against that and move pending things to a different story.

mmulholla:
@baiju In the new code build.yml exits early if sanity check fails but then prepare_pr_comment.py looks for messages from sanity check. I think previously we eliminated that early exit and checked more often it passed.

baijum:
> @baijum In the new code build.yml exits early if sanity check fails but then prepare_pr_comment.py looks for messages from sanity check. I think previously we eliminated that early exit and checked more often it passed.

Before every exit from the sanity check, an error message has been set with a print statement like this:
`::set-output name=sanity-error-message::...`

The only possible scenario where the message not set is when installing dependencies fails.
For the last few hours, it's happening as there is an ongoing infra failure: https://status.python.org/

.github/workflows/build.yml
```diff
@@ -140,6 +143,8 @@ jobs:
 
       - name: 'Prepare PR comment'
         if: always()
+        env:
+          SANITY_ERROR_MESSAGE: ${{ steps.sanity_check_pr_content.outputs.sanity-error-message }}
```
mmulholla, line 147: We set a variable here for a sanity check error but if there is a sanity check error we would never get here because of the  "Reflect on Sanity Check PR Content" step which exits on sanity check failure well before we get here.

baijum: There is an `if: always()` condition (line. 145), so this is going to run irrespective of any other previous job fails or not.

mmulholla: So "exit 1" doesn't exit, perfectly obvious! Got it.

dperaza4dustbit (approved):

dperaza4dustbit:
Before we merge this, lets make sure there is not regression with an end to end submission test.

mmulholla (approved):

PR #187, opened 2021-05-27
Add job to create token for testing

baijum:
This workflow can be triggered by anyone who has write access to the
Git repository.  The job produce a token and it can be accessed from a
pastebin which is going to be expired after the first access.

Jira: https://issues.redhat.com/browse/HELM-187

### DIFF
--- a/.github/workflows/token.yml
+++ b/.github/workflows/token.yml
@@ -1,0 +1,46 @@
+name: Generate Token
+
+on:
+  workflow_dispatch: {}
+
+jobs:
+  token-for-testing:
+    runs-on: ubuntu-20.04
+    steps:
+
+      - name: Checkout
+        uses: actions/checkout@v2
+
+      - name: Set up Python 3.x
+        uses: actions/setup-python@v2
+        with:
+          python-version: '3.9'
+
+      - name: Get Date
+        id: get-date
+        run: |
+          echo "::set-output name=date::$(/bin/date -u "+%Y%m%d")"
+        shell: bash
+
+      - uses: actions/cache@v2
+        id: cache
+        with:
+          path: oc
+          key: ${{ steps.get-date.outputs.date }}
+
+      - name: Install oc
+        if: ${{ steps.cache.outputs.cache-hit != 'true' }}
+        run: |
+          curl -sLO https://mirror.openshift.com/pub/openshift-v4/clients/ocp/stable/openshift-client-linux.tar.gz
+          tar zxvf openshift-client-linux.tar.gz oc
+
+      - name: Create token
+        run: |
+          python3 -m venv ve1
+          cd scripts && ../ve1/bin/pip3 install coolname && cd ..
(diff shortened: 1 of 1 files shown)


.github/workflows/token.yml
```diff
@@ -0,0 +1,46 @@
...
+          curl -sLO https://mirror.openshift.com/pub/openshift-v4/clients/ocp/stable/openshift-client-linux.tar.gz
+          tar zxvf openshift-client-linux.tar.gz oc
+
+      - name: Create token
+        run: |
+          python3 -m venv ve1
+          cd scripts && ../ve1/bin/pip3 install coolname && cd ..
+          cd scripts && ../ve1/bin/pip3 install -r requirements.txt && cd ..
+          cd scripts && ../ve1/bin/python3 setup.py install && cd ..
+          ./oc login --token=${{ secrets.CLUSTER_TOKEN }} --server=https://api.ocpappsvc-osd.zn6c.p1.openshiftapps.com:6443
```
dperaza4dustbit, line 43: Is this going to allow anyone to create access token into our test cluster? I would not open this to the world. Since we have to maintain that cluster healthy to be able to run the service. If someone has their own cluster they can follow instruction to create their own SA and token there and use from their fork.

dperaza4dustbit (changes requested):

.github/workflows/token.yml
baijum: @dperaza4dustbit As I mentioned in the description, this workflow can be triggered (manually) by anyone who got write access to the Git repository. This would be helpful to debug issues directly in the same cluster where we are running the checks.

dperaza4dustbit: How do we provide write access? What is the policy there.

baijum: I think that's going to be our team members and probably others on a need basis. cc. @pedjak

dperaza4dustbit: OK, so what I understand is then that only folks that are explicitly added to have write access to our repo through secrets.CLUSTER_TOKEN? To make it cleaner does it make sense that we add a new role/group just for ocp_jobs access. Running a job is no necessarily write access.

dperaza4dustbit: And then only repo admins can add or remove access to members that need to debug something. It could even be temp access.

baijum: Who runs a [workflow_dispatch job](https://docs.github.com/en/actions/managing-workflow-runs/manually-running-a-workflow) is a GitHub restriction. If necessary, we could create an allow-list inside the script and check who triggered the job.

baijum: I have limited users who can run this job here: 188730b962c16613263aea08afde5ff979aa275e

baijum:
@dperaza4dustbit If no other concerns, please approve!

.github/workflows/token.yml
dperaza4dustbit: OK

dperaza4dustbit (approved):

PR #181, opened 2021-05-20
Update instruction to run tests from the fork

baijum:
* Access to an OpenShift cluster is mandatory to run chart-testing
* Personal Access Token (PAT) is not required anymore
* Make instruction to generate token better with more details and links

PR #195, opened 2021-06-09
Remove hard-coded API server value

baijum:
The API_SERVER GitHub secret should be created with a base64 encoded
string (Warning: That means that's not really a secret -- the value will
be visible in the log)

To generate base64 encoded string for a URL:

  echo -n "https://api.ocpappsvc-osd.zn6c.p1.openshiftapps.com:6443" | base64

Background: This issue came up while updating the documentation about how to run the workflow in the partners fork connecting to their own cluster. API server was hard-coded. This PR moves the API Server URL to GitHub secrets. See this PR: https://github.com/openshift-helm-charts/charts/pull/181

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -116,12 +116,13 @@ jobs:
         env:
           KUBECONFIG: /tmp/ci-kubeconfig
         run: |
+          API_SERVER=$( echo -n ${{ secrets.API_SERVER }} | base64 -d)
           gpg --version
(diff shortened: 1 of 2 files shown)


baijum:
I have tested this PR here: https://github.com/miles2go/repo/pull/116

mmulholla (approved):
LGTM

dperaza4dustbit (approved):

PR #199, opened 2021-06-14
Add empty index

### LINKED ISSUE #198: https://charts.openshift.io/index.yaml file should return empty YAML rather then 404 error when repo is empty
Currently, in case `https://charts.openshift.io` repo is empty, then serving `https://charts.openshift.io/index.yaml` index file returns a `404` error.

The repo should return existing file with a valid empty YAML content such as :
```yaml
[]
```
or
```yaml
--- []
```

Being an empty repository is a corner case but still a valid one IMHO and so should be properly reflected in the repository's index.

Returning 404 error while asking for index in case of the empty repository might cause problems at client side (such as OpenShift Dev Console)

baijum:
Fixes #198

### DIFF
--- a/index.yaml
+++ b/index.yaml
@@ -1,0 +1,3 @@
+apiVersion: v1
+entries: {}
+generated: '2021-06-14T11:45:34.694098+00:00'


baijum:
Tested it here: https://github.com/miles2go/repo/pull/117

mmulholla (approved):

PR #206, opened 2021-06-22
build.yml: use base64 encoded github secret for CLUSTER_TOKEN

zonggen:
Consider using base64 encoded github secret for CLUSTER_TOKEN. Because
currently API_SERVER is encoded, it might be better to stay consistent
while using secrets.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -121,7 +121,7 @@ jobs:
           docker pull quay.io/redhat-certification/chart-verifier:latest
           curl https://raw.githubusercontent.com/helm/helm/master/scripts/get-helm-3 | bash
           if [ "${{steps.sanity_check_pr_content.outputs.report-exists}}" != "true" ]; then
-            ./oc login --token=${{ secrets.CLUSTER_TOKEN }} --server=${API_SERVER}
+            ./oc login --token=$( echo -n ${{ secrets.CLUSTER_TOKEN }} | base64 -d) --server=${API_SERVER}
             ve1/bin/sa-for-chart-testing --create chart-verifier-ci-${{ steps.get-repository.outputs.repository }}-${{ github.event.number }} --token token.txt --server ${API_SERVER}
           fi
           cd pr-branch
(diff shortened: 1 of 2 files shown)


baijum:
> Consider using base64 encoded github secret for CLUSTER_TOKEN.

This is dangerous :-)

We should not expose our CLUSTER_TOKEN to the world.

But the API URL is not really a secret, that's why it's base64 encoded.

zonggen:
We are not exposing the secrets, and from my understanding base64 encoding is not for encryption (please correct me if this is incorrect). The idea of this PR is to keep secrets consistent in terms of encoding, since during development I had to remind myself one is encoded and the other one is not.

This PR is a just a thought during development :)

baijum:
> base64 encoding is not for encryption

That is the reason why API URL is encoded and kept in GitHub secret (Yes, we abused it to avoid hard-coding in the script).
As a result of this, the value of the API URL is visible in the GitHub Action log.

But that is not the case with "CLUSTER_TOKEN" - this is a real secret that needs to be protected. It should not appear anywhere in the log. If we use base64 enoding, the value will be visible to anyone who opens the GitHub Actions log.

zonggen:
Thanks for the explanation. Though I've seen both secrets as ` API_SERVER=$( echo -n *** | base64 -d)` and `./oc login --insecure-skip-tls-verify --token=*** --server=${API_SERVER}` under GitHub Actions log during testing and both secrets are redacted by default. As I'm confused by why encoding can change the visibility, am I missing something?

baijum:
Please see this line in the console log:
https://github.com/openshift-helm-charts/charts/pull/211/checks#step:14:64

```
Logged into "https://api.ocpappsvc-osd.zn6c.p1.openshiftapps.com:6443" as "system:serviceaccount:prod-chart-verifier-infra:chart-verifier-admin" using the token provided.
```
It shows the API URL in plain text without masking.

baijum:
Not only that particular line. The same URL is visible in many places in that log.

zonggen:
I agree, those urls are printed by `oc` commands if I'm not mistaken and we should dump the output if we really don't want to expose the url. But regarding the encoding as proposed by this PR, I'm not sure why it would affect the visibility as CLUSTER_TOKEN is already used and redacted in the workflow.

baijum:
> I'm not sure why it would affect the visibility as CLUSTER_TOKEN is already used and redacted in the workflow.

CLUSTER_TOKEN is a secret and that should be treated that way. Through base64 encoding, we are losing the protection offered by GitHub. If we accidentally or through some error traceback, if that value gets logged it will be visible to all. Let's avoid that possibility.

zonggen:
As per the discussion, closing this PR as it adds additional security concern.

PR #208, opened 2021-06-22
make sure sanity check finds a match for report only submission

mmulholla:
sanity check logic fails for a report only submission because it fails to find a valid file in the chart directory.  This is because faulty logic meant  it did not register the report as a valid file.

After fixing this logic I re-submitted a report only submission with a timestamp updated and it reran the build process and published the chart.


baijum (approved):
LGTM

PR #210, opened 2021-06-22
Don't try to create a release for report only submission

mmulholla:
Add a check for the tag being non blank before running the release. It is not set for report only submission and release will always fail if it is not set.

Tested report only and report with chart. Creates release in latter case.


baijum (approved):
This change is not necessary, but good to check.
LGTM

PR #214, opened 2021-06-25
Adding sha256 digest for each of the existing charts.

dperaza4dustbit:
We are adding sha256 digest as an additional chart field. This matches
what helm index command does and it improves our capabilities to do
integrity checking. In Addition this will fix a gui dependency on digest
to use as a unique ID for the charts. A different PR will be sent to
compute this digest as part of the ci flow.

### DIFF
--- a/index.yaml
+++ b/index.yaml
@@ -11,6 +11,7 @@ entries:
     apiVersion: v2
     appVersion: 1.7.2
     description: Official HashiCorp Vault Chart
+    digest: b07be2a554ecbe6a6dd48ea763ed568de317d17cf1a19fb11ddb562983286555
     home: https://www.vaultproject.io
     icon: https://github.com/hashicorp/vault/raw/f22d202cde2018f9455dec755118a9b84586e082/Vault_PrimaryLogo_Black.png
     keywords:
@@ -41,6 +42,7 @@ entries:
     apiVersion: v2
     appVersion: 1.7.3
     description: Official HashiCorp Vault Chart
+    digest: 97e274069d9d3d028903610a3f9fca892b2620f0a334de6215ec5f962328586f
     home: https://www.vaultproject.io
     icon: https://github.com/hashicorp/vault/raw/f22d202cde2018f9455dec755118a9b84586e082/Vault_PrimaryLogo_Black.png
     keywords:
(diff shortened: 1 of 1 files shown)


mmulholla:
@dperaza4dustbit should the digests not be prefixed by "sha256:"

mmulholla:
For ibm-object-storage-plugin-2.1.2.tgz the digest are different:

https://github.com/IBM/charts/blob/master/repo/ibm-helm/ibm-object-storage-plugin-2.1.2.tgz?raw=true
06efa1e26f8a7ba93a6e6136650b0624af2558cc44a4588198fca322f9219e32

https://github.com/IBM/charts/blob/master/repo/ibm-helm/ibm-object-storage-plugin-2.1.2.tgz
3e11de48a194466de8b6c82f2ffcea453e561b8a797120b3bd29f0013e6aaeb4

should we use the  latter?

dperaza4dustbit:
> For ibm-object-storage-plugin-2.1.2.tgz the digest are different:
> 
> https://github.com/IBM/charts/blob/master/repo/ibm-helm/ibm-object-storage-plugin-2.1.2.tgz?raw=true
> 06efa1e26f8a7ba93a6e6136650b0624af2558cc44a4588198fca322f9219e32
> 
> https://github.com/IBM/charts/blob/master/repo/ibm-helm/ibm-object-storage-plugin-2.1.2.tgz
> 3e11de48a194466de8b6c82f2ffcea453e561b8a797120b3bd29f0013e6aaeb4
> 
> should we use the latter?

No, raw one is the one we should do since is the one that gives us the binary package first one if you do a get is and html response. Also, raw is included in the original URL and what people will check for integrity

dperaza4dustbit:
> @dperaza4dustbit should the digests not be prefixed by "sha256:"

What do you think? If you run helm index it does not add it, so I think is understood, I just can't find the documentation. Searching the repo now. So this is a maybe for me, don't want to be too distinct from what people are used to.

zonggen (approved):
All digest matches, LGTM!

mmulholla:
thx @dperaza4dustbit all looks good.

mmulholla (approved):
LGTM

zonggen:
Just checked adding the "sha256:" prefix works too.

PR #216, opened 2021-06-30
[WIP] workflows: add redhat and community workflow logic

zonggen:
As part of supporting different vendor types for openshift helm charts, this adds the github actions for redhat and community submission. For design details please refer to: [docs/workflow-build.md](https://github.com/zonggen/charts/blob/redhat-and-community-workflow/docs/workflow-build.md).

TODO:
- [ ] Modify `vendorType=redhat` to `vendorType=community` when we manually publish failed redhat chart (needs multi-profile support from chart-verifier: https://github.com/redhat-certification/chart-verifier/pull/161). We can handle this case by using a environment variable / flag in `chartprreview.py` during `chart-certification-redhat-community` job.

NOTE:
 - Did not use the `labeler` action in triage stage because the timing of the tag addition is unknown and the process seems to be asynchronous to the main workflow. For example, if the labeler was triggered to add a `redhat` label and when the following job was looking for a vendorType label (`redhat` in this case), it could not find it because at that time the label might not have been added yet (not serial process).

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -2,12 +2,15 @@ name: CI
 
 on:
   pull_request_target:
-    types: [opened, synchronize, reopened, edited, ready_for_review]
+    types: [opened, synchronize, reopened, edited, ready_for_review, labeled]
 
 jobs:
   triage:
     name: Triage
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-20.04
+    if: |
+      github.event.action != 'labeled' ||
+      github.event.label.name == 'publish'
     steps:
       - name: Checkout
         uses: actions/checkout@v2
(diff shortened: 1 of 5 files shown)


zonggen:
Next step would be to add a trigger to the label `publish` to allow maintainers to manual merge and release the submitted PR by adding the label `publish`.

zonggen:
I think for now we are stuck with partial duplicates between different flows because GitHub action only support composite run steps with shell scripts, not with `uses` actions ([example usage](https://docs.github.com/en/actions/creating-actions/creating-a-composite-run-steps-action#creating-an-action-metadata-file)). This matters to us since the duplicates contain both `run` and `uses` steps.

However, the GitHub Action folks seem to be working on the `uses` support ([discussion](https://docs.github.com/en/actions/creating-actions/creating-a-composite-run-steps-action#creating-an-action-metadata-file)), so the duplicates will likely be cleaned up sooner.

.github/workflows/build.yml
```diff
@@ -115,22 +377,36 @@ jobs:
         id: verify_pr
         env:
           KUBECONFIG: /tmp/ci-kubeconfig
+          VENDOR_TYPE: ${{ needs.triage.outputs.category }}
         run: |
           API_SERVER=$( echo -n ${{ secrets.API_SERVER }} | base64 -d)
           gpg --version
-          docker pull quay.io/redhat-certification/chart-verifier:latest
+          docker pull quay.io/redhat-certification/chart-verifier:main
```
dperaza4dustbit, line 384: Lets make sure we change this back once we are done testing

.github/workflows/build.yml
```diff
@@ -219,8 +495,437 @@ jobs:
...
+        uses: softprops/action-gh-release@master
+        continue-on-error: true
+        with:
+          tag_name: ${{ steps.release-charts.outputs.tag }}
+          files: pr-branch/report.yaml
+          fail_on_unmatched_files: true
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
+  chart-certification-redhat-community:
```
dperaza4dustbit, line 522: This is not what I was expecting. Single flow for redhat and in the code that generates the index.yaml we change given a flag we set here. The difference between this flow and community is that we continue even when report fails but take into account report to set the charts.openshift.io/providerType: https://github.com/openshift-helm-charts/charts/blob/main/scripts/src/chartrepomanager/chartrepomanager.py#L145
Something like if category == redhat and report_result = failed: providerType = community

dperaza4dustbit (changes requested):
Not done, reviewing setting this up for a test on my fork

.github/workflows/build.yml
zonggen: Originally I tried to add a single `community` job and made it depend on `triage` and `redhat`, then it got skipped everytime when `redhat` job got skipped (in the case where it was a community chart in the first place). I will confirm again because I am fairly certain that something didn't work fundamentally.

zonggen:
Closing this PR in favor of https://github.com/openshift-helm-charts/charts/pull/224.

PR #220, opened 2021-07-06
Added reports for ent and pro charts

ntinvo:

PR #221, opened 2021-07-06
chartprreview: strip username obtained from github api

zonggen:
Related: https://github.com/openshift-helm-charts/charts/pull/220

### DIFF
--- a/scripts/src/chartprreview/chartprreview.py
+++ b/scripts/src/chartprreview/chartprreview.py
@@ -46,7 +46,7 @@ def verify_user(directory, username, category, organization, chart):
 
     data = open(owners_path).read()
     out = yaml.load(data, Loader=Loader)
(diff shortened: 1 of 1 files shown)


scripts/src/chartprreview/chartprreview.py
```diff
@@ -46,7 +46,7 @@ def verify_user(directory, username, category, organization, chart):
 
     data = open(owners_path).read()
     out = yaml.load(data, Loader=Loader)
-    if username not in [x['githubUsername'] for x in out['users']]:
+    if username.strip() not in [x['githubUsername'] for x in out['users']]:
```
dperaza4dustbit, line 49: I don't think we need this, right?

zonggen: yep closing, for reference, users should create pull request against main, not their test branch. The original issue was because in the target branch, the username is not in the OWNERS file.

PR #229, opened 2021-07-08
No need to pass msg to get_vendor_type function

dperaza4dustbit:
The msg var is set inside the function right before it exits, don't see the need to pass at all.

This fixes issue with the flow: https://github.com/dperaza4dustbit/charts/runs/3023503083?check_suite_focus=true


zonggen:
LGTM

mmulholla (commented):
LGTM

PR #252, opened 2021-07-15
add chart

rohammosalli:


rohammosalli:
Hello, The `Missing mandatory check : v1.0/contains-test` for this step pass in my system with  chart-verifier but in  in your CI fails 

`docker run -it --rm quay.io/redhat-certification/chart-verifier:latest verify https://github.com/actano/charts-2/blob/main/charts/partners/actano/webclient-01/0.1.0/charts/webclient-01-0.1.0.tgz\?raw\=true`

dperaza4dustbit:
Thanks for your submission @rohammosalli we are in the process of fixing a few messaging bugs while analyzing the report. The messages you see are a bit misleading and we already have a fix that we are taking through our pipeline. We welcome you feedback given your experience with the helm certification flow. Having said that you have a few things that need to be corrected for this chart to be certified:

1. The submission is not following the different options documented, you are mixing two ways to submit. One way we support is Chart Package (.tgz) + report. Another way is source + report. In this PR you have both and the source is not in the right directory. Please peak one method to submit and resubmit the PR. Use this as reference https://github.com/openshift-helm-charts/charts/tree/main/docs
2. I also ran the checks as you suggested and get same report back you do. The root issues you have right now are the the images used are not Red Hat certified and that the test cases do not pass. Once you have your Chart Test working and your Red Hat images certified, you can resubmit.

rohammosalli:
Hello @dperaza4dustbit , thank you so much, for Image I used `registry.access.redhat.com/ubi8/nodejs-12` as our base image, is not enough to be Red Hat images certified? or there is some other steps to get Red Hat images certified?

dperaza4dustbit:
> Hello @dperaza4dustbit , thank you so much, for Image I used `registry.access.redhat.com/ubi8/nodejs-12` as our base image, is not enough to be Red Hat images certified? or there is some other steps to get Red Hat images certified?

Base image for certification is not enough cause you are adding layers on top of the base with your image. Looks like this is the default used in the Values file:

repository: eu.gcr.io/allex-artifacts/rplan-webclient-redhat@sha256:06ce398859ca5e9f7018d5c3994439e525f3d6600ee9b3ce91847eafb7bbf547

For A chart to be certified it is a pre-requisite to use images that are also certified. Reach out to your RedHat Partnership contact get instructions on that or connect you with our certification support team if you have questions. Let me know if you need help finding out information of container image certification: https://connect.redhat.com/en/partner-with-us/red-hat-container-certification

PR #253, opened 2021-07-16
Fixing Red Hat path with pkg only and failed chart test

dperaza4dustbit:
The version check will be skipped if the report has failures but we did not exit.


zonggen (approved):
LGTM

mmulholla (approved):
LGTM

PR #254, opened 2021-07-19
Use report command

mmulholla:
The PR contains:

1. Update to use new chart verifier report command to get from the report
- check results
- chart digests 
- chart-uri 
- chart metadata
- vendor-type and profile version
- annotations
2. Added a global environment varaible for the verifier image so that when the image used is changed it only need to be updated in one place. 
3. Added some INFO message outputs at the start of python functions including interesting input parameter values. 

Tested chart source only, chart tarball only, report only and chart + report.

This addresses:
- https://issues.redhat.com/browse/HELM-198

However because the chart contrnt is now  obtained as JSON instead of yaml it also addresses:
- https://issues.redhat.com/browse/HELM-210

### DIFF
--- a/scripts/requirements.txt
+++ b/scripts/requirements.txt
@@ -5,3 +5,5 @@ PyYAML==5.4.1
 requests==2.25.1
 semver==2.13.0
 urllib3==1.26.5
+docker
+six
--- a/scripts/src/chartprreview/chartprreview.py
+++ b/scripts/src/chartprreview/chartprreview.py
@@ -15,6 +15,9 @@ try:
 except ImportError:
     from yaml import Loader, Dumper
 
+sys.path.append('../')
+from report import report_info
+
 def write_error_log(directory, *msg):
     with open(os.path.join(directory, "errors"), "w") as fd:
         for line in msg:
@@ -101,12 +104,12 @@ def verify_signature(directory, category, organization, chart, version):
 
 def match_checksum(directory, category, organization, chart, version):
     submitted_report_path = os.path.join("charts", category, organization, chart, version, "report.yaml")
-    submitted_report = yaml.load(open(submitted_report_path), Loader=Loader)
-    submitted_digest = submitted_report["metadata"]["tool"]["digest"]
+    submitted_digests = report_info.getReportDigests(submitted_report_path)
+    submitted_digest = submitted_digests["chart"]
 
     generated_report_path = "report.yaml"
-    generated_report = yaml.load(open(generated_report_path), Loader=Loader)
-    generated_digest = generated_report["metadata"]["tool"]["digest"]
+    generated_digests = report_info.getReportDigests(generated_report_path)
+    generated_digest = generated_digests]["chart"]
 
     if  submitted_digest != generated_digest:
         msg = f"[ERROR] Digest is not matching: {submitted_digest}, {generated_digest}"
@@ -114,8 +117,7 @@ def match_checksum(directory, category, organization, chart, version):
         sys.exit(1)
 
 def check_url(directory, report_path):
-    report = yaml.load(open(report_path), Loader=Loader)
-    chart_url = report["metadata"]["tool"]['chart-uri']
+    chart_url = report_info.getReportChartUrl(report_path)
 
     try:
         r = requests.head(chart_url)
@@ -149,9 +151,9 @@ def check_url(directory, report_path):
 def match_name_and_version(directory, category, organization, chart, version):
     submitted_report_path = os.path.join("charts", category, organization, chart, version, "report.yaml")
     if os.path.exists(submitted_report_path):
-        submitted_report = yaml.load(open(submitted_report_path), Loader=Loader)
-        submitted_report_chart_name = submitted_report["metadata"]["chart"]["name"]
-        submitted_report_chart_version = submitted_report["metadata"]["chart"]["version"]
+        submitted_report_chart = report_info.getReportChart(submitted_report_path)
+        submitted_report_chart_name = submitted_report_chart["name"]
+        submitted_report_chart_version = submitted_report_chart["version"]
 
         if submitted_report_chart_name != chart:
             msg = f"[ERROR] Chart name ({submitted_report_chart_name}) doesn't match the directory structure (charts/{category}/{organization}/{chart}/{version})"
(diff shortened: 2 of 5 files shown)


scripts/src/report/report_info.py
```diff
@@ -0,0 +1,86 @@
...
+    from yaml import CLoader as Loader, CDumper as Dumper
+except ImportError:
+    from yaml import Loader, Dumper
+
+REPORT_ANNOTATIONS = "annotations"
+REPORT_RESULTS = "results"
+REPORT_DIGESTS = "digests"
+REPORT_METADATA= "metadata"
+
+def getReportInfo(report_path,info_type,profile_type,profile_version):
```
dperaza4dustbit, line 18: use the pythonic way to name methods: https://www.python.org/dev/peps/pep-0008/#method-names-and-instance-variables, I would run this through a linter like this one https://code.visualstudio.com/docs/python/linting

dperaza4dustbit: space after comas in signature

scripts/src/report/report_info.py
```diff
@@ -0,0 +1,86 @@
...
+
+    docker_command = "report " + info_type + " charts/"+os.path.basename(report_path)
+
+    set_values = ""
+    if profile_type:
+        set_values = "profile.vendortype="+profile_type
```
dperaza4dustbit, line 24: Avoid concatenation like this use the % operator or the string.format(). It is more efficient.

dperaza4dustbit: I personally like something like this:
set_values = "profile.vendortype=%s" % profile_type

scripts/src/report/report_info.py
```diff
@@ -0,0 +1,86 @@
...
+    if profile_version:
+        if set_values:
+            set_values = set_values + ","
+        set_values = set_values + "profile.version="+profile_version
+
+    if set_values:
+        docker_command += " --set " + set_values
```
dperaza4dustbit, line 31: be consistent above you do x = x + b and here you use x += b. But if you switch to string formatting it would be something like this:
docker_command = "%s --set %s" % (docker_command, set_values)

scripts/src/report/report_info.py
```diff
@@ -0,0 +1,86 @@
...
+
+    client = docker.from_env()
+    report_directory = os.path.dirname(os.path.abspath(report_path))
+    output = client.containers.run(os.environ.get("VERIFIER_IMAGE"),docker_command,stdin_open=True,tty=True,stderr=True,volumes={report_directory: {'bind': '/charts/', 'mode': 'rw'}})
```
dperaza4dustbit, line 35: nice

zonggen (commented):
Some ideas when we are doing cleanups:
 * can use a formatter like `autopep8` to format code before submit changes
 * can use a logger like `logging` package and reduce use of `print`'s
 * single single quotes for string literals and use double quotes for error messages: https://docs.ckan.org/en/ckan-2.7.3/contributing/python.html#:~:text=Use single quotes,ve got an error!".
 * add some docstring documentations on what each script is doing

dperaza4dustbit (commented):
Logic is sound, just minor python specific standards. A linter will help you with little things like that while you can focus on the logic. Once we add the linter as a gate we will have that too.

scripts/src/report/report_info.py
```diff
@@ -0,0 +1,86 @@
...
+        for report_annotation in report_out[REPORT_ANNOTATIONS]:
+            annotations[report_annotation["name"]] = report_annotation["value"]
+
+        return annotations
+
+    return report_out[info_type]
+
+
+def getReportAnnotations(report_path):
+    annotations = getReportInfo(report_path,REPORT_ANNOTATIONS,"","")
```
zonggen, line 53: might be better to call `chart-verifier report all` with the corresponding profile once and parse the result, instead of calling the docker command from each of the functions.

scripts/src/chartprreview/chartprreview.py
```diff
@@ -52,6 +56,7 @@ def get_modified_charts(directory, api_url):
     sys.exit(1)
 
 def verify_user(directory, username, category, organization, chart):
+    print("[INFO] Verify user.",username, category, organization, chart)
```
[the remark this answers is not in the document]
zonggen: or f-strings which is relatively new and Baiju already uses in the scripts: 
* https://github.com/openshift-helm-charts/charts/blob/main/scripts/src/sanitycheckpr/sanitycheckpr.py#L58
* https://realpython.com/python-f-strings/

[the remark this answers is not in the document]
dperaza4dustbit: that works too yes.

zonggen:
opened https://github.com/openshift-helm-charts/charts/issues/255 for cleanup proposal

mmulholla:
> Some ideas when we are doing cleanups:
> 
>     * can use a formatter like `autopep8` to format code before submit changes
> 
>     * can use a logger like `logging` package and reduce use of `print`'s
> 
>     * single single quotes for string literals and use double quotes for error messages: https://docs.ckan.org/en/ckan-2.7.3/contributing/python.html#:~:text=Use single quotes,ve got an error!".
> 
>     * add some docstring documentations on what each script is doing

Agree on all counts for when we restructure the workflow sand scripts.

scripts/requirements.txt
```diff
@@ -5,3 +5,5 @@ PyYAML==5.4.1
 requests==2.25.1
 semver==2.13.0
 urllib3==1.26.5
+docker
```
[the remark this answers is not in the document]
mmulholla: Added.

scripts/src/chartprreview/chartprreview.py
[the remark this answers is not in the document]
mmulholla: I changed to how David suggested. Can change again when we restrusture if needed.

scripts/src/report/report_info.py
mmulholla: Fixed all my new functions.

scripts/src/report/report_info.py
mmulholla: Done.

scripts/src/report/report_info.py
mmulholla: done.

scripts/src/report/report_info.py
mmulholla: Would then need ti store the results to serve different requests, not sure how to do that in python.

zonggen: I was thinking about setting a private variable in `report_info.py` and checks if this variable is empty each `getReportInfo` run and skip docker run if the info is already retrieved. But then I realized some runs contain vendor type and version, it might require setting vendor type & version in the first run. I don't disagree with the current approach because i don't think the overhead will be large.

mmulholla: OK, lets leave as-is for now, and consider changing in the future.

dperaza4dustbit (approved):

zonggen (approved):

PR #257, opened 2021-07-21
check all files in a PR

mmulholla:
Fix problem with limit on how many files the github api returns. Default is 30.

Code now gets a 100 files at a time, if 100 files are returned from the previous request the next page is requested.

Need to do some more testing tomorrow but wanted to get some early eyes on the fix.

### DIFF
--- a/scripts/src/sanitycheckpr/sanitycheckpr.py
+++ b/scripts/src/sanitycheckpr/sanitycheckpr.py
@@ -26,29 +26,39 @@ def ensure_only_chart_is_modified(api_url, repository, branch):
     r = requests.get(files_api_url, headers=headers)
     pattern = re.compile(r"charts/"+TYPE_MATCH_EXPRESSION+"/([\w-]+)/([\w-]+)/([\w\.-]+)/.*")
     reportpattern = re.compile(r"charts/"+TYPE_MATCH_EXPRESSION+"/([\w-]+)/([\w-]+)/([\w\.-]+)/report.yaml")
-    count = 0
-    match_found = False
-    for f in r.json():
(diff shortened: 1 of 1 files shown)


mmulholla:
all set now, testing complete.
Tested: PR 245 gets a failure
Tested logic with 99,100,101 files in a PR with and without an OWNERS file.
Tested also with 1 file.

scripts/src/sanitycheckpr/sanitycheckpr.py
```diff
@@ -26,29 +26,39 @@ def ensure_only_chart_is_modified(api_url, repository, branch):
...
-                print("::set-output name=report-exists::true")
-            if not match_found:
-                pattern_match = match
-                match_found = True
-            elif pattern_match.groups() != match.groups():
-                msg = f"[ERROR] PR must only include one chart"
+    page_number = 1
+    max_page_size,page_size = 100,100
+
+    while page_size == max_page_size:
```
dperaza4dustbit, line 32: Does the response has a bool indicating it has more pages? What happens in the case where there are exactly 100 files and you try to get page 2? Would it return empty list or fail? That will make a diff in the logic

dperaza4dustbit (commented):
Looks good just one edge case question

scripts/src/sanitycheckpr/sanitycheckpr.py
zonggen: In that case it will return 200 with empty list: https://github.com/openshift-helm-charts/charts/pull/257#discussion_r674000540

dperaza4dustbit (approved):

PR #266, opened 2021-07-27
Python script for test workflow

mmulholla:
Script to use for testing workflow. Checks if the test should run based on requirements

- manual invocation (no PR)
   - User is defined as an approver in the repo OWNERS file.
- automatic invocation (PR)
   - User is defined as an approver in the repo OWNERS file.
   - All changes are to the workflow or tests. All files must be in one or more of:
        - .gitgub/workflow/
        - scripts/
        - tests/

workflow to use the script will follow

### DIFF
--- a/scripts/src/workflowtesting/checkprforci.py
+++ b/scripts/src/workflowtesting/checkprforci.py
@@ -21,10 +21,8 @@ def check_if_ci_only_is_modified(api_url):
     max_page_size,page_size = 100,100
 
     workflow_found = False
-    other_found = False
-    test_config_path = ""
 
-    while (page_size == max_page_size) & (not other_found):
+    while (page_size == max_page_size):
 
         files_api_query = f'{files_api_url}?per_page={page_size}&page={page_number}'
         r = requests.get(files_api_query,headers=headers)
@@ -42,11 +40,8 @@ def check_if_ci_only_is_modified(api_url):
             elif pattern_test.match(filename):
                 workflow_found = True
             else:
-                other_found = True
-                break
+                return False
 
-    if other_found:
-        return False
 
     return workflow_found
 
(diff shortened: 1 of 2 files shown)


scripts/src/workflowtesting/checkprforci.py
```diff
@@ -0,0 +1,84 @@
...
+    pattern_script = re.compile(r"scripts/.*")
+    pattern_test = re.compile(r"tests/.*")
+    page_number = 1
+    max_page_size,page_size = 100,100
+
+    workflow_found = False
+    other_found = False
+    test_config_path = ""
+
+    while (page_size == max_page_size) & (not other_found):
```
zonggen, line 27: Should this be `and` instead of the bitwise `&`?

scripts/src/workflowtesting/checkprforci.py
```diff
@@ -0,0 +1,84 @@
...
+            filename = f["filename"]
+            if pattern_workflow.match(filename):
+                workflow_found = True
+            elif pattern_script.match(filename):
+                workflow_found = True
+            elif pattern_test.match(filename):
+                workflow_found = True
+            else:
+                other_found = True
+                break
```
zonggen, line 46: Instead of break, we can early return here..

scripts/src/workflowtesting/checkprforci.py
```diff
@@ -0,0 +1,84 @@
...
+
+def main():
+    parser = argparse.ArgumentParser()
+    parser.add_argument("-u", "--api-url", dest="api_url", type=str, required=False,
+                                        help="API URL for the pull request")
+    parser.add_argument("-n", "--verify-user", dest="username", type=str, required=True,
+                        help="check if the user can run tests")
+    args = parser.parse_args()
+    if verify_user(args.username):
+        if not args.api_url or check_if_ci_only_is_modified(args.api_url):
```
zonggen, line 79: Could you add a comment to note this checks `not (PR submission and contain other changes)`? I find it hard to understand without the brackets..

scripts/src/workflowtesting/checkprforci.py
mmulholla: second condition removed - no longer needed - see below.

scripts/src/workflowtesting/checkprforci.py
mmulholla: changed to early return - so second condition removed from while condition above.

scripts/src/workflowtesting/checkprforci.py
mmulholla: I revamped this section after testing to improve messaging. New version should be clear.

zonggen (approved):

dperaza4dustbit (approved):

PR #274, opened 2021-07-27
add Ziti Hosting Endpoint chart from NetFoundry

qrkourier:


dperaza4dustbit:
Thanks for submitting this chart @qrkourier, we welcome any feedback you can give us given your experience with the certification flow. It looks like you are missing a step to create partner project. Follow these instructions: https://github.com/openshift-helm-charts/charts/tree/main/docs#preparation. I you have question follow the support link above.

PR #277, opened 2021-07-29
tests: cleanup extensive environment variables

zonggen:

### DIFF
--- a/tests/functional/notifier.py
+++ b/tests/functional/notifier.py
@@ -59,8 +59,6 @@ def _make_gihub_request(method, uri, body=None, params={}, headers={}, verbose=F
     return resp_json
 
 # Call this method directly if you are not creating a verification issue nor a version change issue.
-
-
 def create_an_issue(title, description, assignees=[], labels=[]):
     uri = "issues"
     method = "post"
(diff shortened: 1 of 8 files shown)


tests/functional/utils.py
```diff
@@ -43,7 +49,7 @@ def get_run_result(secrets, run_id):
     return run['conclusion']
 
 
-@retry(stop_max_delay=15_000)
+@retry(stop_max_delay=15_000, wait_fixed=1000)
```
zonggen, line 52: Hit GitHub request limit while testing, set this interval period to slow down the requests.

tests/functional/notifier.py
```diff
@@ -94,6 +92,9 @@ def create_verification_issue(chart_name, chart_owners, report_url, software_nam
     access_token -- An optional github access token secret. If not passed will try to get from GITHUB_AUTH_TOKEN environment variable\n
     """
 
+    if os.environ.get('NOTIFY_ID'):
```
zonggen, line 95: Use this `NOTIFY_ID` if provided to override the chart_owners to tag

dperaza4dustbit (approved):
I think we can improve these test module to reduce the duplicate logic like:
if not bot_token:
        bot_name = "github-actions[bot]"
        ...

tests/functional/notifier.py
mmulholla: The notify id could be "OWNERS", "none" or a githib id. I would expect to see some logic depending in its value, like reading the OWNERS file if set to OWNERS. Also if "none" there is no id to notify.

mmulholla:
Also I get the impression there is a lot of duplicated code. Need to look into making some common functions.

PR #280, opened 2021-08-04
tests: fix chart src testing when src dir already exists

zonggen:
When there's already a `src` directory under the target location under `charts/`,
     the test case will fail because the code will _move_ the test chart
     _into_ the `src` directory, which will result in mismatched SHA value.

This fixes the issue by removing the src directory in the runner and
test with the test charts.


zonggen:
Before fix: https://github.com/abai-test-bot/sandbox/runs/3242615408?check_suite_focus=true
After fix: https://github.com/abai-test-bot/charts/runs/3243467802?check_suite_focus=true

mmulholla:
The code changes appear identical, can we make the code a single function?

zonggen:
Refactored the code and tested

mmulholla (approved):
LGTM

dperaza4dustbit (approved):
lgtm

PR #282, opened 2021-08-09
workflow: use pull_request_target to get secrets

zonggen:
Previously we used `pull_reqeust` as event target but GitHub does not
allow secret access on pull_request because of security reasons:
https://docs.github.com/en/actions/reference/encrypted-secrets#using-encrypted-secrets-in-a-workflow.

This fix uses `pull_request_target` to access secrets and use the
bot_token.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,7 +1,7 @@
 name: Test Workflow
 
 on:
-  pull_request:
+  pull_request_target:
     types: [opened, synchronize, reopened, edited]
   workflow_dispatch:
     inputs:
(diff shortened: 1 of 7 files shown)


.github/workflows/test.yml
```diff
@@ -36,6 +36,8 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v2
         with:
+          repository: ${{ github.event.pull_request.head.repo.full_name }}
+          ref: ${{ github.event.pull_request.head.sha }}
```
zonggen, line 40: Checkout to pull request repo and HEAD.

dperaza4dustbit (approved):
lgtm

PR #283, opened 2021-08-09
simple change just to test triggered tests

dperaza4dustbit:
Should see tests triggered agains sandbox


zonggen:
The permissions are read from OWNERS file approvers.

PR #284, opened 2021-08-09
workflow: handle case when it's workflow only change but user is not authorized

zonggen:

### DIFF
--- a/scripts/src/workflowtesting/checkprforci.py
+++ b/scripts/src/workflowtesting/checkprforci.py
@@ -4,6 +4,8 @@ import os
 import requests
 import json
 import yaml
+import sys
+
 try:
     from yaml import CLoader as Loader, CDumper as Dumper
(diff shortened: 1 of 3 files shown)


zonggen:
Need to investigate further to see if it will have more side effects, will life WIP when ready.

scripts/src/workflowtesting/checkprforci.py
```diff
@@ -76,14 +78,17 @@ def main():
             print(f"::set-output name=run-tests::true")
         else:
             print(f"[INFO] User not authorized for manual invocation - do not run tests.")
+            sys.exit(1)
```
dperaza4dustbit, line 81: One way you can do this better is to raise an exception the will fail and print the message

zonggen:
Ready for review.

dperaza4dustbit (approved):

PR #305, opened 2021-08-24
Add Infinispan chart

ryanemerson:


zonggen:
Thank you for submitting the chart! The script was expecting the chart name to be `infinispan-helm-charts`. Therefore the path should be
```diff
- charts/community/infinispan/infinispan/0.1.0
+ charts/community/infinispan/infinispan-helm-charts/0.1.0
```
and the tarball to be
```diff
- infinispan-0.1.0.tgz
+ infinispan-helm-charts-0.1.0.tgz
```

The reason behind this is the script is respecting the name defined in the `Chart.yaml` which is `infinispan-helm-charts`. Running `helm package infinispan-helm-charts/` also generates `infinispan-helm-charts-0.1.0.tgz`.

ryanemerson:
@zonggen Thanks for explaining. I have updated the chart name to simply "infinispan".

dperaza4dustbit:
Hello @ryanemerson do you want to take a look at why the testing is failing?

ryanemerson:
> Hello @ryanemerson do you want to take a look at why the testing is failing?

@dperaza4dustbit The cause isn't obvious to me. AFAICT the only checks that failed are ones that have `type: Optional`.

dperaza4dustbit:
@ryanemerson yeah it is not blocking the PR just want to give you a chance to debug before it make the repo.

ryanemerson:
I think the failures are fine:

- `images-are-certified` - It's upstream so we can't fix this.
- `has-kubeversion` - The Chart should work with all Openshift versions that have Helm chart support. I will make a note to add this in the future though.
- `chart-testing` - I don't know what this is TBH, any ideas?

dperaza4dustbit (approved):

dperaza4dustbit:
@ryanemerson we fixed an issue in the publishing flow, if you rebase I can publish this

dperaza4dustbit:
> I think the failures are fine:
> 
> * `images-are-certified` - It's upstream so we can't fix this.
> * `has-kubeversion` - The Chart should work with all Openshift versions that have Helm chart support. I will make a note to add this in the future though.
> * `chart-testing` - I don't know what this is TBH, any ideas?

You can add test cases to a chart the can be executed after a chart has been installed: https://helm.sh/docs/topics/chart_tests/

PR #310, opened 2021-08-25
wildfly: mv from charts/redhat to charts/community

zonggen:
This has come up recently in the discussion,

> The main attribute for community are charts that are supported by a
community and not covered by enterprise support. Community charts may
come from Red Hatters, from partners (i.e. in their own upstream work).
Charts that are in placed under Red Hat should be commercially
supported, i.e. because they provide a supported installation method for
a product or component. Certification only applies to products from
partners, not from Red Hat. - Hugo Rivero


dperaza4dustbit (approved):
Owners file will have to change to add co

PR #319, opened 2021-09-01
Add eap74 1.1.0 chart

jmesnil:

# Status of the chart verifier:

All checks pass except the following:

```
  - check: v1.0/has-kubeversion
    type: Mandatory
    outcome: FAIL
    reason: Kubernetes version is not specified
```
-> We don't have any hard-dependency on a Kube version

```
  - check: v1.0/images-are-certified
    type: Mandatory
    outcome: FAIL
    reason: 'Failed to certify images : Failed to get images, error running helm template
        : execution error at (eap74/templates/buildconfig-s2i-build-artifacts.yaml:2:4):
        Git source repository URL is required'
```
-> this helm chart requires user input (to locate their application Git repo), it can not be run as is

```
  - check: v1.0/contains-test
    type: Mandatory
    outcome: FAIL
    reason: Chart test files do not exist
```
-> likewise, this helm chart requires user input (to locate their application Git repo), it can not be run as is for testing

```
  - check: v1.0/chart-testing
    type: Mandatory
    outcome: FAIL
    reason: |-
        Error running process: executing helm with args "install eap74-rh9g9lumgn /root/.cache/chart-verifier/_charts_1_1_0_eap74_1_1_0_tgz/eap74 --namespace default --wait --values /tmp/chart-testing-668347137/values.yaml": exit status 1
        ---
        Error: Kubernetes cluster unreachable: Get "http://localhost:8080/version?timeout=32s": dial tcp 127.0.0.1:8080: connect: connection refused
```
-> again, this helm chart requires user input (to locate their application Git repo), it can not be run as is for testing


dperaza4dustbit:
@jmesnil main reason is not merging automatically is this: 
PR should only modify chart related files: charts/redhat/redhat/eap74/OWNERS
Move the change OWNERs to other PR. The report for Red Hat charts are not blocking this from merging it is a guidance for you to make the category Red Hat in the OpenShift Console. If these checks fail it will still merge and release, but they will be marked Community in the index file and therefore in the console. The same standards we follow with partners is the same we follow for Red Hat charts to be able to categorize as Red Hat. See the charts under partners folder for an example of how they got those checks to pass. Vault one is one that provide the source as well like you do here. BTW if you provide the source, not need to run the report yourself we will run it for you if you don't send the report.yaml that is more for environment where you need specific config to run your tests.

jmesnil:
@dperaza4dustbit I've split the OWNERS update in a separate #321 

We are planning to publish only tgz in that repo as the development of the charts is already done in our repo at https://github.com/jbossas/eap-charts/.
I prefer copying a tgz that has passed our own testing rather than duplicated the charts source code here.

Having said that, we need to improve our testing as now this does not align with Helm test since our charts are not usable out of the box.

dperaza4dustbit:
@jmesnil you can send tgz only as well, understand your usecase

dperaza4dustbit:
@jmesnil you are now hitting an issue that looks like a bug on our end, we are investigating root cause. This chart should be able to merge and publish

charts/redhat/redhat/eap74/1.1.0/report.yaml
```diff
@@ -0,0 +1,95 @@
+2021/09/01 14:46:21 [INFO] Missing required value: Git source repository URL is required
+RenderManifests error : execution error at (eap74/templates/buildconfig-s2i-build-artifacts.yaml:2:4): Git source repository URL is required
```
dperaza4dustbit, line 2: These two lines above are not needed in this report.yaml

dperaza4dustbit (changes requested):

charts/redhat/redhat/eap74/1.1.0/report.yaml
dperaza4dustbit: That is what is causing the failure

PR #322, opened 2021-09-01
Fixing community path with force-publish label

dperaza4dustbit:
Need to skip version check when verifying community charts.


mmulholla (approved):
LGTM

zonggen (approved):
LGTM!

PR #330, opened 2021-09-15
Fixing semver regular expression

dperaza4dustbit:
We found a bug on our regular expression no covering a version format we should support.


mmulholla (approved):

zonggen (approved):
LGTM

PR #332, opened 2021-09-16
do not build non code changes

mmulholla:
Update workflows to not build and exit with success if PR only includes:

- README.md
- docs/*.md
- release/release_info.json

This also ensure that tests are run if a PR contains workflow changes and a change matching one of these criteria.

This is required in the chart repository ahead of creating the first auto release of workflows form the development rrepository/

### DIFF
--- a/scripts/src/workflowtesting/checkprforci.py
+++ b/scripts/src/workflowtesting/checkprforci.py
@@ -19,10 +19,14 @@ def check_if_ci_only_is_modified(api_url):
     pattern_workflow = re.compile(r".github/workflows/.*")
     pattern_script = re.compile(r"scripts/.*")
     pattern_test = re.compile(r"tests/.*")
+    pattern_release = re.compile(r"release/release_info.json")
+    pattern_readme = re.compile(r"README.md")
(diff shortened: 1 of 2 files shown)


scripts/src/workflowtesting/checkprforci.py
```diff
@@ -19,10 +19,14 @@ def check_if_ci_only_is_modified(api_url):
     pattern_workflow = re.compile(r".github/workflows/.*")
     pattern_script = re.compile(r"scripts/.*")
     pattern_test = re.compile(r"tests/.*")
+    pattern_release = re.compile(r"release/release_info.json")
+    pattern_readme = re.compile(r"README.md")
+    pattern_docs = re.compile(r"docs/([\w-]+)\.md")
```
zonggen, line 24: I would prefer putting these patterns in an array and iterating on them instead of numerating, e.g. `workflow_files = [r".github/workflows/.*", r"scripts/.*", r"tests/.*"]` and `do_not_build_files = [r"release/release_info.json", r"README.md", r"docs/([\w-]+)\.md"]`. Then iterate on them in the while loop instead of repeating. Please adjust the names to your preference since I couldn't come up with a good name.

PR #360, opened 2021-10-20
Add Helm Chart for WildFly 25

jmesnil:


dperaza4dustbit:
@jmesnil maybe then use this chart to update the annotations too, and send to the wildly path instead of the redhat path

PR #361, opened 2021-10-20
move WildFly charts under community/wildfly

jmesnil:
update its OWNERS list


jmesnil:
@dperaza4dustbit thanks. I've amended this PR to only include the move of the Owners File

I will have to make a new release of the Helm Chart to add the 2 annotation `charts.openshift.io/name: Wildfly, charts.openshift.io/provider: Wildfly` and will open a PR later with it

PR #367, opened 2021-10-26
Add Helm Chart for WildFly 25

jmesnil:


dperaza4dustbit:
@jmesnil In the path charts/community/redhat/wildfly/1.5.2/report.yaml the vendor redhat should match the label here: https://github.com/openshift-helm-charts/charts/blob/main/charts/community/redhat/wildfly/OWNERS#L9. This is a test we make during the certification flow to make sure vendor is under the right directory. In this case since we are not changing the vendor path to avoid the duplicate issue we have spoke about, we have to keep the label in the OWNERs file to be redhat. However, you are still able to change the vendor name here: https://github.com/openshift-helm-charts/charts/blob/main/charts/community/redhat/wildfly/OWNERS#L10. This is how I see we move forward for now until we figure out how to merge two entries in our index.yaml to avoid duplicates in this cases:

1. Change you OWNERs file to have label redhat in another PR
2. Rebase this PR once the OWNERs change is in and let it run the checks again
3. I will review and publish the new version of the charts

jmesnil:
> 1. Change you OWNERs file to have label redhat in another PR

Done in https://github.com/openshift-helm-charts/charts/pull/372

dperaza4dustbit:
@jmesnil should be able to rebase and let me know to do the publish

jmesnil:
@dperaza4dustbit I've rebased my PR. thanks

dperaza4dustbit:
PR was merged but release failed, https://github.com/openshift-helm-charts/charts/runs/4260095048?check_suite_focus=true

dperaza4dustbit:
we are checking on this

jmesnil:
The Helm Chart does not specify a `kubeVersion` (as we don't have a strong requirement for any Kubernetes version).
It was already the case for the existing chart though.

dperaza4dustbit:
Yes we broke the publish part it with a new feature that convert kubeVersion to OpenShiftVersions for a support statement. We are adding more testing and doing you publishing manually. Will let you know when wildfly is published. On the kubeVersion will be best practice in the long run to add. You can do ranges like kubeVersion >= 1.20.0 which includes multiple OCP versions. Up to you but that is one of the check we verify.

dperaza4dustbit:
fyi @jmesnil https://github.com/openshift-helm-charts/charts/pull/379 should fix the index to add your 1.5.2 release. Once that merges you should be good to go

PR #378, opened 2021-11-19
Fix community and redhat path when kubeVersion not set in chart

dperaza4dustbit:
We are not checking if kubeVersion is in chart before trying to
get the value. This is causing redhat and comunity charts that
do not have kubeVersion set to fail on release step.

Signed-off-by: David Peraza <dperaza@redhat.com>


mmulholla (approved):
LGTM

PR #388, opened 2022-01-11
make submission requirements clearer

mmulholla:
Added info to submission requirements to address: https://issues.redhat.com/browse/HELM-223

see also: https://github.com/openshift-helm-charts/development/pull/93

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -66,10 +66,19 @@ You can submit chart related changes in three methods:
 1. Submit a chart without Chart Verification Report
 
    - Chart as a tarball
+      - must be created using ```helm package```
+      - name is ```<chart-name>-<chart-version>.tgz```
    - Chart in a directory
 
 2. Submit a Chart Verification Report without the Chart
+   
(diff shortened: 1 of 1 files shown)


docs/README.md
```diff
@@ -66,11 +66,20 @@ You can submit chart related changes in three methods:
 1. Submit a chart without Chart Verification Report
 
    - Chart as a tarball
+      - must be created using ```helm package```
```
zonggen, line 69: I think this should be "**can** be created using `helm package`". A user can also do `tar -cvzf <target>.tgz <chart-directory>`.

tisutisu: I think 'helm package' command should be recommended as by default the resulting tar file is of format <chart-name>-<chart-version>.tgz .While for tar command user have to specify the target tgz file name, which is prone to user error.

mmulholla: @zonggen @tisutisu have update based on your suggestions.

zonggen (approved):

tisutisu (approved):
LGTM

PR #389, opened 2022-01-11
Add Helm Chart for WildFly 26

jmesnil:


jmesnil:
I'm not sure how you are reviewing the chart manually but if that can helps, the diff to the last incorporated chart for WildFly is https://github.com/wildfly/wildfly-charts/compare/wildfly-1.5.2...wildfly-1.5.4.

TL;DR: We have linted our Helm templates, added a `kubeVersion` field and updated the chart to use the WildFly `26.0` images by default.

dperaza4dustbit:
Merging this one, but notice that other than the certified image, that is not required for community charts, you are only missing the helm test part, if you add that on next version this will be a good chart to show as an example for other community submitter that maybe want to become RedHat partners.

jmesnil:
thanks!

PR #391, opened 2022-01-13
updated chart uri for nodered v0.0.2

prajyot-narulkar925:


index.yaml
```diff
@@ -368,7 +346,7 @@ entries:
     name: nodered-chart
     type: application
     urls:
-    - https://github.com/rhm-samples/edge-helm-charts/blob/master/nodered-chart-0.0.2.tgz
+    - https://github.com/rhm-samples/edge-helm-charts/blob/master/nodered-chart-0.0.2.tgz?raw=true
```
dperaza4dustbit, line 349: This change is correct, but does not match what the report for 0.0.2 version has, lets remove this change here and send another PR for 0.0.3 version with the new url in the report.

index.yaml
```diff
@@ -326,28 +326,6 @@ entries:
...
-    appVersion: 1.16.0
-    description: A Helm chart for NodeRed.
-    digest: 8bc2576c5c07648ea377e68cbc5151e3c7799d1563256ebe571772e86e3e3746
-    icon: https://raw.githubusercontent.com/rhm-samples/node-red-operator/node-red-logo/_images/nodered-logo.png
-    kubeVersion: '>=1.18.0'
-    name: nodered-chart
-    type: application
-    urls:
-    - https://github.com/rhm-samples/edge-helm-charts/blob/master/nodered-chart-0.0.1.tgz?raw=true
-    version: 0.0.1
```
dperaza4dustbit, line 350: I would only remove a version like this if it is something bad like a security issue or something serious like that, if it merely does not work, I would not worry too much about have older versions of a chart that does not work.

dperaza4dustbit (changes requested):

komish:
Closing this PR due to its age. Please re-submit if you wish to make further changes.

PR #400, opened 2022-01-18
Add a Vert.x chart

jponge:


jponge:
That workflow does not make any sense.

When I add `charts/redhat/redhat/vertx/OWNERS` as part of the pull-request then checks failed because it should be submitted separately (why?).

When I remove that file from the PR then the same checks complain that it does not exist.

jponge:
So I moved back the file

mmulholla:
@jponge 

Please submit a PR containing just the OWNERS file. The PR will require maintainer review so we can check there is nothing awry. Once that is merged toy can submit the PR with the chart.

jponge:
@mmulholla done, see #428

PR #408, opened 2022-02-03
Adding Spring Boot REST HTTP example to charts.openshift.io

Sgitario:
Repo: https://github.com/snowdrop/rest-http-example


Sgitario:
@baijum any help with this? 
If I commit the OWNERS file, the workflow complains because it does exist.
If I do not commit the OWNERS file, the workflow also complains because it does not exist.

PR #566, opened 2022-09-23
Delete charts/partners/exfo/u-verifier-chart directory

arasyid-aziz:


komish:
Deleting a partner chart is more involved than simply deleting these resources from this repository. We'll close this issue out, and evaluate your project internally, as it corresponds with a given certification project on the Red Hat side.

PR #605, opened 2022-11-16
ehrbase enterprise helm chart certification

bjoernpauli:
ehrbase enterprise helm chart certification


mmulholla:
@bjoernpauli I see that you used the provider delivery control feature for this PR which means that the chart will not be published in the openshift catalogue. I wanted to make sure that this was your intention because providerDelivery control was set to true in the OWNERS file due to a bug in OWNERS file creation and my concern is you set it in the report to get the PR to pass.

PR #649, opened 2022-12-13
remove mobileum unwanted charts

mmulholla:


elhadjici:
I confirm that all charts  in the PR must need to be delete  except **charts/partners/mobileum/caddy-v015/OWNERS**.
for the two charts we plan to update (with right providerDelivery value) below also I think you can just delete them as we are creating  new chart from scratch for all their helm chart
https://github.com/openshift-helm-charts/charts/blob/main/charts/partners/mobileum/keycloak/OWNERS
https://github.com/openshift-helm-charts/charts/tree/main/charts/partners/mobileum/caddy

PR #679, opened 2023-01-10
add report.yaml for chart certification

zeeshanAt:


komish:
Going to go ahead and close this as stale, given that there is a conflicting report already in place.

PR #701, opened 2023-01-16
Fix partner type v partners directory in metrics

mmulholla:
Quick fix for metrics failure.


Kartikey-star (approved):
LGTM

mmulholla:
Although tests failed they were not connected to this fix. For tests that worked the workflow shows the metrics running ok. for example: https://github.com/openshift-helm-charts/sandbox/pull/52140

PR #705, opened 2023-01-18
add SupportedOCPVersion annotation to all latest charts without it

mmulholla:
Some older charts were missing the  charts.openshift.io/supportedOpenShiftVersions which is now being used by the developer console to provide information about a chart. The annotation was added in chart verifier version 1.4.0 so charts using an earlier version are missing the attribute.


Kartikey-star (approved):
LGTM

PR #740, opened 2023-02-16
Set providerDelivery to true for Mavenir SMSF

rdavid:


komish:
From what I can tell, the project associated with this PR appears to be archived on the Red Hat side. There appears to be a replacement that has been published. To that end, I'll close this out. If you believe I've done so in error, feel free to re-submit and we can take a look further.

PR #805, opened 2023-05-03
change pr bot condition to return true only if all flows passed

### LINKED ISSUE #785: Fix issue with workflow that causes the bot to post a success message when failing in an unexpected place.
Ex. https://github.com/openshift-helm-charts/charts/pull/599#issuecomment-1516608443

itroyano:
The bot would verify if all results are a success and only then return a success message.
failing a step that isn't covered specifically in the `elif`, would produce a generic error message with additional output if possible.

Fixes #785


komish:
Closing in favor of https://github.com/openshift-helm-charts/development/pull/223 which is this PR, but in the proper place to roll out to the production workflows.

PR #815, opened 2023-05-09
Add MySQL persistent helm chart

phracek:
This pull request adds mysql persistent storage Helm Chart

Mysql Helm chart was tested in OpenShift 4 environment here: https://github.com/sclorg/helm-charts/pull/30

and results are:

```bash
test_mysql_imagestreams.py::TestHelmMySQLImageStreams::test_package_imagestream[8.0-el9-registry.redhat.io/rhel9/mysql-80:latest] [32mPASSED[0m[32m [ 25%][0m
test_mysql_imagestreams.py::TestHelmMySQLImageStreams::test_package_imagestream[8.0-el8-registry.redhat.io/rhel8/mysql-80:latest] [32mPASSED[0m[32m [ 50%][0m
test_mysql_imagestreams.py::TestHelmMySQLImageStreams::test_package_imagestream[8.0-el7-registry.redhat.io/rhscl/mysql-80-rhel7:latest] [32mPASSED[0m[32m [ 75%][0m
test_mysql_template.py::TestHelmMySQLDBPersistent::test_package_persistent [32mPASSED[0m[32m [100%][0m
```

### DIFF
--- a/charts/redhat/redhat/mysql-persistent/0.0.1/src/templates/deploymentconfig.yaml
+++ b/charts/redhat/redhat/mysql-persistent/0.0.1/src/templates/deploymentconfig.yaml
@@ -1,0 +1,84 @@
+apiVersion: apps.openshift.io/v1
+kind: DeploymentConfig
+metadata:
+  annotations:
+    template.alpha.openshift.io/wait-for-ready: "true"
+  labels:
+    template: mysql-persistent-template
+  name: {{ .Values.database_service_name }}
+spec:
+  replicas: 1
+  selector:
+    name: {{ .Values.database_service_name }}
+  strategy:
+    type: Recreate
+  template:
+    metadata:
+      labels:
+        name: {{ .Values.database_service_name }}
+    spec:
+      containers:
+      - env:
+        - name: MYSQL_USER
+          valueFrom:
+            secretKeyRef:
+              key: database-user
+              name: {{ .Values.database_service_name }}
+        - name: MYSQL_PASSWORD
+          valueFrom:
+            secretKeyRef:
+              key: database-password
+              name: {{ .Values.database_service_name }}
+        - name: MYSQL_ROOT_PASSWORD
+          valueFrom:
+            secretKeyRef:
+              key: database-root-password
+              name: {{ .Values.database_service_name }}
+        - name: MYSQL_DATABASE
+          valueFrom:
+            secretKeyRef:
+              key: database-name
+              name: {{ .Values.database_service_name }}
+        image: ' '
+        imagePullPolicy: IfNotPresent
+        livenessProbe:
+          exec:
+            command:
+            - /bin/sh
+            - -i
+            - -c
+            - MYSQL_PWD="$MYSQL_PASSWORD" mysqladmin -u $MYSQL_USER ping
+          initialDelaySeconds: 30
+          timeoutSeconds: 1
+        name: mysql
+        ports:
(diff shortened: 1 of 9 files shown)


phracek:
Can you please retrigger tests.

komish:
@phracek Rerunning tests here.

komish:
I'm not sure what's going on here, but it would seem that actions are having trouble as well, so I'll re-run this once githb status is healthy. https://www.githubstatus.com/

phracek:
Would it be possible to merge it?

komish:
@phracek I need to take a closer look at the actions. Will do so as soon as I get a chance.

komish:
@phracek in my testing, this chart outright hangs chart-verifier. Looking further, and will update with more as I have it.

komish:
Updating here - it seems that the DeploymentConfig's empty image value is throwing our certified image check for a loop. Looking into why, and what a reasonable path forward should be.

charts/redhat/redhat/mysql-persistent/0.0.1/src/templates/deploymentconfig.yaml
```diff
@@ -0,0 +1,84 @@
...
+          valueFrom:
+            secretKeyRef:
+              key: database-root-password
+              name: {{ .Values.database_service_name }}
+        - name: MYSQL_DATABASE
+          valueFrom:
+            secretKeyRef:
+              key: database-name
+              name: {{ .Values.database_service_name }}
+        image: ' '
```
komish, line 42: I think you should be able to remove this key. Having it be an empty string is causing an issue with the image certification tests.

I'm not overly familiar with DeploymentConfig resources, but I'm understanding that this will get populated anyway by the configuratoin in the .spec.trigger section. Correct me if I'm wrong. 

Either way, I was able to get past ImagesAreCertified by removing this key locally.

charts/redhat/redhat/mysql-persistent/0.0.1/src/templates/deploymentconfig.yaml
```diff
@@ -0,0 +1,84 @@
...
+        persistentVolumeClaim:
+          claimName: {{ .Values.database_service_name }}
+  triggers:
+  - imageChangeParams:
+      automatic: true
+      containerNames:
+      - mysql
+      from:
+        kind: ImageStreamTag
+        name: mysql:{{ .Values.mysql_version }}
```
komish, line 82: I think you need to add the `namespace` key with `openshift` as the value here. This is causing chart-testing's installation to fail because the DeploymentConfig and ImageStreamTag resources are never ready.

komish:
@phracek check out the comments above, I believe these were, at least in part, responsible for the failures. Let me know when this is resolved. If it doesn't auto-merge (because of a success case), then feel free to ping me to make sure I see it.

FWIW, we're working on resolving the bot's misbehavior. It's not catching the failed verification state and so you see a success message posted, but the merge never runs (because the workflow didn't actually succeed).

charts/redhat/redhat/mysql-persistent/0.0.1/src/templates/deploymentconfig.yaml
komish: Also worth mentioning that chart-testing failed for me altogether even with this fix that I deployed locally. I'm not sure if you're expecting that to succeed or not.

PR #840, opened 2023-05-11
add canvas-platform helm charts

gurolakman:
- add canvas-platform helm charts
- add canvas-platform helm charts
- add canvas-platform helm charts
- add canvas-platform helm charts


komish:
@gurolakman We're considering the additional permissions required here.

For the time being, if you'd like to submit a report from executing chart-verifier against your own cluster, you can submit that and provide it to us which should unblock you here.

gurolakman:
can you help me w/ submitting a report? i've never done it so far so i don't know what exactly to specify for <report-uri> when executing the chart-verifier as follows:

# chart-verifier report <report-uri>

my understanding is that once i can successfully run the above w/ a proper <report-uri> argument, it should generate a *.yaml file which i should submit together w/ my helm package. 

i'm guessing i'd have to close this pull request & initiate a new one to submit a package w/ the report. i may have to specify the partner connect support ticket w/ the new PR #. pls confirm my understanding is accurate. 

last but not the least, we have a bunch of similar helm packages for other products; so, it'll be important to find a lasting solution or workaround for certifying all of them. pls let us know what that will be. thanks much!

komish:
You always have the option of submitting the report along with your chart. The goal here is to allow you to run the certification tooling against your own cluster (instead of ours).

You shouldn't need to open a new PR, just adjust the contents of this PR and re-submit it. Resubmission will retrigger the test automation on our end. As as result, you shouldn't need to change your Connect profile for this project.

What you're looking to execute is `chart-verifier verify` against your chart tarball. We'll do the report checking on our end when you resubmit.

An example of what your directory structure might look like is [here](https://github.com/openshift-helm-charts/charts/tree/main/charts/partners/castai/castai-agent/0.52.0).

As mentioned, this path just helps you self-unblock while we resolve the access issues that we're seeing with your chart, which uses a hostPath mount. We're discussing internally. Just wanted to give you another option.

gurolakman:
i just did this:

$ git commit -m "add report.yaml" 
[main 87356fab] add report.yaml
 1 file changed, 105 insertions(+)
 create mode 100644 charts/partners/telenity/canvas-platform/3.1.5/report.yaml
$ gh pr create
? Where should we push the 'main' branch? gurolakman/charts
a pull request for branch "gurolakman:main" into branch "main" already exists:
https://github.com/openshift-helm-charts/charts/pull/840

does it look right or am i missing something? thanks for your help!

gurolakman:
my earlier attempt must have not worked. given the following:

$ git remote --verbose
origin	git@github.com:gurolakman/charts.git (fetch)
origin	git@github.com:gurolakman/charts.git (push)
upstream	git@github.com:openshift-helm-charts/charts.git (fetch)
upstream	git@github.com:openshift-helm-charts/charts.git (push)

could you help me w/ the instructions to use to re-submit this pr. thanks in advance!

komish:
This can vary wildly depending on your setup, but I think what you would want to do is something like

```
git push origin main
```

This is based on the idea that you're using your `main` branch for this PR, and you've added an additional commit which means you shouldn't have to force push.

gurolakman:
i just executed:

$ git push origin main
Enter passphrase for key '/root/.ssh/id_ed25519': 
Enumerating objects: 14, done.
Counting objects: 100% (14/14), done.
Delta compression using up to 8 threads
Compressing objects: 100% (8/8), done.
Writing objects: 100% (8/8), 1.74 KiB | 1.74 MiB/s, done.
Total 8 (delta 3), reused 0 (delta 0), pack-reused 0
remote: Resolving deltas: 100% (3/3), completed with 3 local objects.
To github.com:gurolakman/charts.git
   930ce62a..87356fab  main -> main

it appears to help re-initiated the checks. so, i'm assuming it did work! we'll see...

gurolakman:
this time, it seems to have failed. due to...

Kube Version >=v1.21.0

i can change this & resubmit. which version should i use?

komish:
The value should be fine, but I'm thinking the format is wrong. 

https://helm.sh/docs/topics/charts/#the-kubeversion-field

I think what you want is something like `'>= 1.21.0'` (no v prefix)

gurolakman:
ok; i changed it in helm package, re-created the report, and re-submitted it

gurolakman:
seems to have worked :-) do you think i should proceed w/ certification of my other helm charts using the same approach (i.e., submitting the report.yaml in addition to the package itself) or wait until you folks finish analyzing & finding a proper solution to the original problem?

komish:
@gurolakman Glad it worked!

You can continue however you see fit, just based on what your timeline looks like. There's nothing wrong with submitting a report alongside your chart. We just offer the opportunity for folks to use our cluster instead of theirs as a convenience, but sometimes special cases have to be taken into consideration such as this one.

Totally up to you! Thanks for your patience thus far.

gurolakman:
thanks to you @komish for all the help. 2 more questions before i let you go ;) 

1. (technical) our image is ubi8 based and we currently use "eth: ens192" in our helm charts for the nic i/f name. is that ok? the reason i'm asking this question is because in our test environment, for example, we have to use "eth: eth0". so, i'm not sure which value to use for most versatility?
2. (non-technical) should i leave the related red hat partner connect support ticket open until you folks find a permanent solution or close it?

once again, thanks!

gurolakman:
re: 1st question

following pls find an excerpt from our helm package:

# cat values.yaml
### Canvas Platform default values.yaml ###
dmid: ## Alias comes from chart.yaml
  port: 2180
  eth: ens192

moagent:
  image: docker.io/gurolakman/platform:3.1
  imagePullPolicy: Always
  port:
    container: 20000
    service: 20000
  replicaCount: 1

the value of the <eth> param above would normally change from one environment to another (e.g., eth0, ens224) where it signifies the network interface name on the k8s node. 

so, my question was in regard to the openshift certification environment in use here. do you know the network interface name we should be using on the openshift [k8s] nodes. that's because if we can't get that name exactly right, our deployment [hence, certification] would eventually fail. i realize we're failing for a different reason now but looking ahead we'd need to tackle this issue as well. so, this is just a heads-up!

PR #847, opened 2023-05-18
add canvas-alarmutils helm charts

gurolakman:


gurolakman:
@komish in order to overcome the HostPath restriction SCC encountered in PR #840, i applied the workaround suggested by https://access.redhat.com/solutions/6560521 on a different helm chart package as part of this PR. unfortunately, this PR also failed during chart install. what i don't know however is whether it failed for the exact same reason (i.e., HostPath restriction) or not. i do realize it also fails for a second test script related reason but that's of less concern at this stage. could you let me know if this PR was able to get over the restriction related issue encountered in PR #840. many thanks!

PR #864, opened 2023-05-23

hasankose86:


hasankose86:
> @hasankose86 Is this a duplicate of #857?

@komish , yes . When i received an error , i am fixing my charts and trying new pull request.

komish:
Understood. You can just push over existing PR/branches. No need to open new ones. With that said, feel free to close whichever one is a duplicate and we can look to resolve any issues you might have with certification.

PR #885, opened 2023-06-10
Delete charts/partners/devnio/nef-nabstract directory

ishanshahnio:


ishanshahnio:
No more required because of abbriviation of the name of partner is wrong "devnio"

PR #895, opened 2023-06-14
Add missing charts.openshift.io/name for eap74 1.1.2

### LINKED ISSUE #894: charts.openshift.io/name annotation not copied when new eap74 release added to index.yaml
I just upgrade the eap74 Helm chart to 1.2.2 in #893 but its name is not properly displayed in OpenShift Dev Console.

It is listed as "Eap74" while the chart itself has the right annotation:

https://github.com/openshift-helm-charts/charts/actions/runs/5269201134/jobs/9526983456#step:23:69

However the index.yaml has not added this annotation  in https://github.com/openshift-helm-charts/charts/commit/42c5188e420f3b9745b4ec2cb7726ab2ea102fc9#diff-08c65c1b5ef8384ef0200558777c52f4110afcbbacc880ac5375d555ca97ee3eR3678-R3684

It seems that the code that is adding some annotations is not copying the `charts.openshift.io/name` from the `Chart.yaml`

jmesnil:
This does not fix #894 but adds back the annotation that was removed when the index.yaml was updated.

### DIFF
--- a/index.yaml
+++ b/index.yaml
@@ -3682,6 +3682,7 @@ entries:
       charts.openshift.io/submissionTimestamp: '2023-06-14T15:41:08.067608+00:00'
       charts.openshift.io/supportedOpenShiftVersions: '>=4.6'
       charts.openshift.io/testedOpenShiftVersion: '4.12'
+      charts.openshift.io/name: JBoss EAP 7.4
     apiVersion: v2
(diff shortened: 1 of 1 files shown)


jmesnil:
@dperaza4dustbit could you please review this PR and merge it if possible?

PR #914, opened 2023-07-03
2.2.16 charts

Bhagyashreek8:


komish:
This PR requires a rebase off of main so that the commit that adds the 2.2.15 chart is no longer considered. Right now, our automation is detecting both charts as being added in this one PR, and we require exactly 1 chart pr PR.

PR #1030, opened 2023-08-28
Add dependencies to git tree.

phracek:


komish:
@phracek your dependencies should not make reference to other charts in-repo in this way. Use helm to pull the dependencies and then make reference to your own local `charts/` directories, I think that's what you're trying to achieve here.

komish:
I'm closing out this PR as stale.

PR #1031, opened 2023-08-28
initial scan report

Digitalpit:
Initial helm chart scan report

### DIFF
--- a/charts/partners/sapiens/idit-runtime-umbrella-chart/1.0.18/report.yaml
+++ b/charts/partners/sapiens/idit-runtime-umbrella-chart/1.0.18/report.yaml
@@ -1,0 +1,128 @@
+apiversion: v1
+kind: verify-report
+metadata:
+    tool:
+        verifier-version: 1.12.1
+        profile:
+            VendorType: partner
+            version: v1.2
+        reportDigest: uint64:10800224655124558046
+        chart-uri: https://idit-artifactory.sapiens.com:443/artifactory/idit-devops-local/chart/idit-runtime/1.0.18/idit-runtime-1.0.18.tgz
+        digests:
+            chart: sha256:4f454ed42420a2b89a4ff3c265415556c1c3bc244fe94522eefa295b33c0cc7b
+            package: 1af6f76aa182df2e0b3436aaff8cefaecd950b7c2d1712b7a9e8faf6e5f4ce73
+        lastCertifiedTimestamp: "2023-08-28T12:44:39.471229+00:00"
+        testedOpenShiftVersion: "4.10"
+        supportedOpenShiftVersions: 4.10 - 4.13
+        webCatalogOnly: false
+    chart:
+        name: idit-runtime
+        home: ""
+        sources: []
+        version: 1.0.18
+        description: A Helm chart for Kubernetes
+        keywords: []
+        maintainers: []
+        icon: ""
+        apiversion: v2
+        condition: ""
+        tags: ""
+        appversion: 1.16.0
+        deprecated: false
+        annotations:
+            app.idit/version: 19.2.0.2
+            charts.openshift.io/name: IDITSuite 18.
+        kubeversion: '>= 1.23.0 <= 1.26.3'
+        dependencies:
+            - name: mono
+              version: 1.0.15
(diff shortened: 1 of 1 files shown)


komish:
Hi there @Digitalpit. The failure you're seeing here seems to be related to a bug in our tooling. I'm working to resolve this, and will send you a ping when this is resolved, as I think I'll need you to regenerate your report using a later version of Chart Verifier.

Nothing to do for you just yet. I'll send a ping when things are in place. Thanks for your patience.

charts/partners/sapiens/idit-runtime-umbrella-chart/1.0.18/report.yaml
```diff
@@ -0,0 +1,128 @@
...
+        icon: ""
+        apiversion: v2
+        condition: ""
+        tags: ""
+        appversion: 1.16.0
+        deprecated: false
+        annotations:
+            app.idit/version: 19.2.0.2
+            charts.openshift.io/name: IDITSuite 18.
+        kubeversion: '>= 1.23.0 <= 1.26.3'
```
komish, line 35: This needs to look like this.

```suggestion
        kubeversion: '>=1.23.0 <=1.26.3'
```

This fix needs to go into your Chart.yaml, and then the report needs to be regenerated using chart-verifier. Submit the updated report, and I think we should get further into the certification process.

I've created an issue to look more closely at our version interpreting logic in the pipeline, but fixing the above should resolve this for you more quickly.

komish:
@Digitalpit workaround listed above ^. Please have a look and we'll look forward to your updated chart.

charts/partners/sapiens/idit-runtime-umbrella-chart/1.0.18/report.yaml
Digitalpit: OK, will try and check

Digitalpit (commented):
Approved

Digitalpit:
Yes, I tried to use this solution and it helped but now there is another error:

Warning:  URL is not accessible: {chart_url} 
403 Client Error: Forbidden for url: https://idit-artifactory.sapiens.com:443/artifactory/idit-devops-local/chart/idit-runtime/1.0.18/idit-runtime-1.0.18.tgz

probably because I used our internal repo to store chart (which is not publicly available - only to whitelisted IPs):
chart-uri: https://idit-artifactory.sapiens.com:443/artifactory/idit-devops-local/chart/idit-runtime/1.0.18/idit-runtime-1.0.18.tgz
But we don't want to publish our chart.
Please advise.

atiwary91:
@komish There registries are private Azure registries where they keep their chart and don't want to share/upload anywhere. Could you suggest what needs to be done here.

komish:
@atiwary91 @Digitalpit

> Partner registries are private Azure registries where they keep their chart and don't want to share/upload anywhere. Could you suggest what needs to be done in this case?

> probably because I used our internal repo to store chart (which is not publicly available - only to whitelisted IPs):
> chart-uri: https://idit-artifactory.sapiens.com:443/artifactory/idit-devops-local/chart/idit-runtime/1.0.18/idit-runtime-1.0.18.tgz
> But we don't want to publish our chart.

Aha, Got it!

In this case, the project is [misconfigured](https://github.com/openshift-helm-charts/charts/blob/main/charts/partners/sapiens/idit-runtime-umbrella-chart/OWNERS#L4). When you created your Connect project, you selected a delivery model indicating that your chart will be distributed through our repos.

You wanted to select "Web Catalog Only", or "Provider Delivered" during your project creation on https://connect.redhat.com, indicating that you will be the sole distributor of your chart. I believe we'll need to close this PR, have you recreate your project indicating that you wish to deliver the charts, and then submit a PR against the new project directory containing your report.

When you run Chart Verifier against your chart, you'll want to use the correct options associated with the "Web Catalog Only" delivery method so that certain chart metadata are stripped out of your report before submission.

https://github.com/redhat-certification/chart-verifier/blob/main/docs/helm-chart-submission.md#web-catalog-only

I'll close this out, but we'll be on the lookout. Any further questions, don't hesitate to update here.

Issue #934, opened 2023-07-19
Vault NetworkPolicy requires additional `from` rules in OCP 4.13


komish:
@tvoran
@jasonodonnell
@tomhjp
@calvn
@swenson

Hi Folks! We've been doing some work to prepare our certification workflows for OCP 4.13, and think you you may need to update and submit a revised chart in prep. You've been pinged as you're listed as owners of the hashicorp/vault charts.

We had issues in our testing with vault's MutatingWebhookConfiguration not being accessible by kube-apiserver in 4.13. Per my testing, this is just because the NetworkPolicy that's created needs an extra match criteria.

The policy looks like this in your latest chart:
https://github.com/openshift-helm-charts/charts/blob/main/charts/partners/hashicorp/vault/0.24.0/src/templates/injector-network-policy.yaml

The `from` items needs to include an additional entry allowing host network workloads to reach it.
```yaml
...
spec:
  podSelector:
    ... removed for brevity ...
  ingress:
    - from:
        - namespaceSelector: {}
        - namespaceSelector:      # add this and the following lines
            matchLabels:
              policy-group.network.openshift.io/host-network: ""
...
```

This is documented here:
https://docs.openshift.com/container-platform/4.13/networking/network_policy/about-network-policy.html#nw-networkpolicy-allow-from-hostnetwork_about-network-policy

We discovered it when chart-testing would run. The webhook would be live and the chart-test would immediately fail because the pod it creates would fail at admission (due to the webhook being inaccessible from the kube-apiserver).

Please let me know how I can help! If it would help, I'm happy to open an issue or a PR to the chart repo if you provide a link.

komish:
Closing this issue in favor of https://github.com/hashicorp/vault-helm/issues/953 which seems to be a better home.

PR #1036, opened 2023-09-01
HashiCorp Vault v0.24.1

tvoran:
https://github.com/hashicorp/vault-helm/releases/tag/v0.24.1


komish:
Hi @tvoran - The issue you're seeing here is related to the included network policy. Please see the solution mentioned here: https://github.com/openshift-helm-charts/charts/issues/934 - this should resolve the problem.

komish:
Closing the loop here, the issue that I reported associated with this chart seemed to be caused by an underlying bug in OCP at the version we used in our certification pipeline. We rolled out an updated Z-stream version and things cleared up.

PR #1047, opened 2023-09-11
Added helm chart verification report

Digitalpit:
Added helm chart verification report

### DIFF
--- a/charts/partners/sapiens/idit-runtime-umbrella/1.0.18/report.yaml
+++ b/charts/partners/sapiens/idit-runtime-umbrella/1.0.18/report.yaml
@@ -6,15 +6,15 @@ metadata:
         profile:
             VendorType: partner
             version: v1.2
-        reportDigest: uint64:16597227322790631861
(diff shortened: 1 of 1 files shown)


Digitalpit:
Chart is stored in our internal repo and it's not accessible for everyone (only whitelisted IPs)
I created this project in RedHat site like this, so should be fine:
image
Please advice

mgoerens:
@Digitalpit In your case, you need to run the report with the `--web-catalog-only` flag. See https://github.com/redhat-certification/chart-verifier/blob/main/docs/helm-chart-submission.md#web-catalog-only

Hope that helps !

Digitalpit:
So I need to create project for helm chart from scratch to be able to fix it?

mgoerens:
Yes, unfortunately those settings cannot be modified, hence you need to create a new Helm Chart certification project.

Please see: https://redhat-connect.gitbook.io/partner-guide-for-red-hat-openshift-and-container/helm-chart-certification/creating-a-helm-chart-certification-project

For the distribution method you'll need to select "Web Catalog Only" as you are hosting the Helm Chart in a private repository.

PR #1076, opened 2023-09-22
add initial OWNERS file for trusted-artifact-signer

sallyom:
Initial PR for onboarding trusted-artifact-signer helm-chart

### DIFF
--- a/charts/redhat/redhat/trusted-artifact-signer/OWNERS
+++ b/charts/redhat/redhat/trusted-artifact-signer/OWNERS
@@ -1,0 +1,13 @@
+chart:
+  name: trusted-artifact-signer
+  shortDescription: A Helm chart for deploying Red Hat Trusted Artifact Signer
+publicPgpKey: null
+users:
+  - githubUsername: JasonPowr
+  - githubUsername: lance 
+  - githubUsername: sabre1041
+  - githubUsername: sallyom
+  - githubUsername: tommyd450
+vendor:
+  label: redhat
+  name: Red Hat


charts/redhat/redhat/trusted-artifact-signer/OWNERS
```diff
@@ -0,0 +1,13 @@
+chart:
+  name: trusted-artifact-signer
```
komish, line 2: Is this a "common name for a tool that exists" or a name that you've picked/generated for this chart? 

If it's the former, does it make sense for us to prefix this with "redhat" as to not squat on an upstream name?

Today, the pipeline prefixes your chart name with your organization (e.g. redhat-trusted-artifact-signer), but in the future we'll be building the index without the organization in the name, and I'm trying to make sure we aren't taking a community name if we can avoid it.

If not, no big deal, we can merge this either way.

sallyom: Thanks for the review! The upstream is `scaffold` (sigstore/scaffold), so this `trusted-artifact-signer` name is unique - however let me ask around on the team to get the preferred chart name - right now we use `scaffolding` in testing. Hold off on a merge for now, I'll report back.

sallyom: @komish we do want to go with `trusted-artifact-signer` . Thanks!

komish (approved):
/lgtm

PR #1119, opened 2023-11-21
add OWNERS file for redhat/knative-istio-authz

Kaustubh-pande:
This is preparation for the initial release of Red Hat Openshift Serverless (knative-istio-authz chart). A chart will be added in a follow-up PR.

The users mentioned in file are Red Hatters.
```
users:
- githubUsername: Kaustubh-pande
- githubUsername: pierDipi
- githubUsername: rudyredhat1
- githubUsername: ReToCode
- githubUsername: mgencur
```

PR #1125, opened 2023-11-23
add redhat prefix in chart name

Kaustubh-pande:
as per comment: https://github.com/openshift-helm-charts/charts/pull/1119#issuecomment-1823428107
add redhat prefix in chart name knative-istio-authz

cc @komish


komish:
No need to worry about this now. In the future, the chart you submit will just need to have the appropriate values. We will guide you through it when the time comes should issues arise.

PR #1136, opened 2023-12-11
chore: use the same checkout configuration in test.yml as in build.yml

nickboldt:
GH action https://github.com/openshift-helm-charts/charts/actions/workflows/test.yml seems to be failing for the last few PR checks. This change might solve that problem?

However I'm not auth'd to run workflow code updates on this repo:

```
  # workflow only change but user not authorized
```

So a code owner will have to run this.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -34,10 +34,8 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v3
         with:
+          ref: ${{ github.event.pull_request.head.ref }}
           repository: ${{ github.event.pull_request.head.repo.full_name }}
-          ref: ${{ github.event.pull_request.head.sha }}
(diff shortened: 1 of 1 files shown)


komish:
Thanks @nickboldt. We have this noted in our development repositories and will roll out the fix in the next release.

komish:
I'll close this out at that time.

komish:
@nickboldt Thanks again. I'm closing this out for now. This isn't quite the fix we want roll out, but the fix we want to roll out is a bit more involved. 

With that said, this issue should go away for contributors if you rebase your fork off of main from this repository. We bumped versions in our GHA runners, which is causing this issue when you send PRs based on older versions of the repo.

The "Workflow Test" action doesn't actually impact certification in any way, but it can still be discouraging for contributors to see a :x: in their action summary.

My apologies for any confusion

PR #1141, opened 2023-12-11
Add .NET imagestreams chart.

tmds:
Fixes https://github.com/redhat-developer/s2i-dotnetcore/issues/477.

cc @omajid @aslicerh


charts/redhat/redhat/redhat-dotnet-imagestreams/0.0.1/src/Chart.yaml
```diff
@@ -0,0 +1,16 @@
+description: |-
+  This content is experimental, do not use it in production. Build and run .NET applications on UBI.
```
tmds, line 2: > This content is experimental, do not use it in production.

I've copied this description from the nodejs chart.

Similar, the _(experimental)_ in the `charts.openshift.io/name`.

PR #1265, opened 2024-03-07
Rename spring-boot-rest-http to redhat-springboot-rest

Sgitario:
As per action requested email from Josh.

### DIFF
--- a/charts/redhat/redhat/redhat-springboot-rest/OWNERS
+++ b/charts/redhat/redhat/redhat-springboot-rest/OWNERS
@@ -1,0 +1,9 @@
+chart:
+  name: redhat-springboot-rest
+  shortDescription: A Helm chart to build and deploy Spring Boot application with REST endpoints
+publicPgpKey: null
+users:
+- githubUsername: Sgitario
+vendor:
+  label: redhat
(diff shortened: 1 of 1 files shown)


Sgitario:
cc @jsm84

jsm84:
@Sgitario The name change looks good, but (something I neglected to mention) as part of the index rollout, we're not fixing old chart releases as we can't "re-release" those.

The fix here is to create a new chart release version, either 0.0.2 or 0.0.1+1 (hyphenated tags won't get picked up by helm as the latest release version). You can leave the 0.0.1 chart release as is.

jsm84:
I should also add that the chart release needs it's own separate PR (from the chart directory and OWNERS file name changes), or I can submit the directory change separately and you can rebase.

Sgitario:
> I should also add that the chart release needs it's own separate PR (from the chart directory and OWNERS file name changes), or I can submit the directory change separately and you can rebase.

So, this PR to add the OWNERS file, and https://github.com/openshift-helm-charts/charts/pull/1273 to add the chart (I will rebase it once this is merged)

jsm84 (approved):
lgtm

PR #1274, opened 2024-03-08
Rename semeru-transition chart to redhat-semeru-transition

levivic:
Rename semeru-transition chart to redhat-semeru-transition


jsm84 (approved):
LGTM. 

Now we just need to open a new release PR to get an updated version of this chart released as v1.0.0+1 (or v1.0.1 since hyphens won't work due to semver restrictions).

I'll go ahead and update the index in the meantime.

PR #1437, opened 2024-05-07
chore: remove bad chart which refers to...

nickboldt:
chore: remove bad chart which refers to non-existant container image

Signed-off-by: Nick Boldt <nboldt@redhat.com>


nickboldt:
closed in favour of https://issues.redhat.com/browse/HELM-531

PR #1527, opened 2024-07-02
Removing nns-docker partner

devopsnns:
Removing nns-docker partner since it does not represent our real entity name and was derived from somewhere, need to name change it to nonamesec in a different PR


mgoerens:
Closing in favor of https://github.com/openshift-helm-charts/charts/pull/1535

PR #1596, opened 2024-09-17
deleted test versions for nonamesec partner charts

itzhak-alayev-nonamesec:


komish:
This kind of operation cannot be executed by a user pull request. Please open a case via the Partner Acceleration Desk to discuss.

PR #1744, opened 2025-03-03
Add model-server chatbot ai sample versions

thepetk:
Adds version `0.1.6` to `model-server-sample` and `chatbot-ai-sample` from `ai-lab` project

PR #1769, opened 2025-03-31
Update `model-server-sample` chart

thepetk:
First of the two new PRs after I've closed https://github.com/openshift-helm-charts/charts/pull/1744

Adds version 0.1.6 to model-server sample of ai-lab project


thepetk:
> For this one, the tooling is telling me that the content of the report isn't aligning with the calculated checksum. Re-run the verifier against this chart and generate a new report, then run `chart-verifier report all <reportfile.yaml>` against it to make sure the content checksum passes.

@komish thanks for the input. the notification was lost somewhere and I just noticed it!

PR #1925, opened 2025-07-02

hasankose86:


mgoerens:
@hasankose86 I see you've been trying to add a new version of your charts. The issue you're hitting is:

> A PR must contain only one chart. Current PR includes files for multiple charts.

Looking at your diff:

image


You should not be removing `i2i5gcorehelm/1.0.0` in this PR, but only should be adding the new version (1.0.9).

Also, note that you should not create a new chart for each version, you should keep everything version under the same chart name. Don't create a `i2i5gcorehelm-v109` chart for your new version, keep it under `i2i5gcorehelm/1.0.9`.

So to summarize, the only changes contained by this PR should be the addition of the 1.0.9 version of your chart under `i2i5gcorehelm/`.

Hope that helps, feel free to ping me if you run into any more issue.

hasankose86:
Hi @mgoerens, actually I tried it the way you suggested at first, but I encountered the same error again. After that, I tried to remove the old charts. I will revert my changes again.

PR #2989, opened 2026-04-16
Add redhat-rhem chart for 1.1.1

keitwb:


keitwb:
@komish I need to be added to the OWNERS file in https://github.com/openshift-helm-charts/charts/pull/2990 and then once we do that will be able to release this without further manual review?

komish:
After that merges, rebase this PR and it should retrigger the workflow. If all checks are green, you should be good to go.

PR #3091, opened 2026-05-14
Replace openbao-helm report-only with chart-only submission

karras:
This should add missing meta data in the OpenShift catalog, and include the chart in the charts repo.


karras:
We'll switch to a chart-only submission in a next release.

PR #3116, opened 2026-05-20
Add chart 5g-upf-global 1.260315.10516330 from Samsung

shlee519:
Helm chart certification for 5g-upf-global version 1.260315.10516330

Partner: Samsung


komish:
Ref:
CERTOPS-2094
CERTOPS-2087

PR #3132, opened 2026-05-26
fix(openbao): change vendor name to 'OpenBao'

pree:


komish:
Please open a partner support ticket for guidance. This kind of change cannot be made here.

pree:
Thanks for the feedback @komish. I've opened up a ticket already earlier today and wasn't sure whether to keep the PR open :)

PR #3146, opened 2026-05-28
Add new owners to Vault chart

schavis:
Add Github users to the owners file to support Red Hat access for submitting OpenShift artifacts


komish:
This must be done via your [Connect portal](https://connect.redhat.com/).

PR #3187, opened 2026-06-09
Add zcm-redis 0.1.0 verification report

Mredward123:
## Chart Information

| Field | Value |
|-------|-------|
| **Chart Name** | zcm-redis |
| **Version** | 0.1.0 |
| **Vendor** | iWhaleCloud |
| **Type** | application |
| **App Version** | aik |
| **Kube Version** | >=1.20.0 |

## Verification Details

| Field | Value |
|-------|-------|
| **Tool Version** | chart-verifier v1.14.1 |
| **Profile** | v1.3 / partner |
| **Tested OpenShift Version** | 4.21 |
| **Supported OpenShift Versions** | >=4.7 |
| **Web Catalog Only** | true |
| **Certified Timestamp** | 2026-06-09T08:01:10+00:00 |

## Verification Results

All Mandatory checks passed: ✅

## Delivery Method

- Web catalog only delivery (no chart tarball submitted)
- Chart is signed, PGP public key registered in OWNERS file

### DIFF
--- a/charts/partners/iwhalecloud/zcm-redis/0.1.0/report.yaml
+++ b/charts/partners/iwhalecloud/zcm-redis/0.1.0/report.yaml
@@ -6,13 +6,13 @@ metadata:
         profile:
             VendorType: partner
             version: v1.3
-        reportDigest: uint64:11506990529405218537
+        reportDigest: uint64:8440614228156750125
         chart-uri: N/A
         digests:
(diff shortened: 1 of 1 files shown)


Mredward123:
The reason for the failure of this test case is that our environment is an offline one and cannot connect to the public network to verify the image. However, in fact, our image has already undergone preflight verification and can be seen on the Red Hat portal. So this error needs to be handled by the backend staff.

komish:
CERTOPS-2114

PR #3269, opened 2026-06-19
Update redist persistent

phracek:
Update version 0.0.3

Use proper version in values.schema.json

enable PVC in Helm chart


mgoerens:
Release job failed due to concurrency issues.

I've retried it and the release was now fully successful.

PR #3428, opened 2026-08-25
Add chart 5g-nrf-global 1.260814.10982825 from Samsung

shlee519:
Helm chart certification for 5g-nrf-global version 1.260814.10982825

Partner: Samsung


komish:
ref: CERTOPS-2171

